PR #12, opened 2018-01-19
Use ctx.which to find nix-build

johnynek:
the current rules assume `nix-build` just works. If it doesn't you get a cryptic message.

I added a call to which. If that fails we print that nix-build was not found. If it is found, we use that path in our call.

I verified that I saw the error message on my machine without nix installed, then installed nix and verified I can then run the tests.

The `which` path was needed for me to be able to run on OSX.


mboes:
Very cool! Did not know about `ctx.which`.

PR #15, opened 2018-04-30
Add build results as indirect GC roots

### LINKED ISSUE #7: Think about garbage collection
rules_nixpkgs has a fundamental problem: neither it nor its users know when the built targets may go away.

Nothing is stopping nix GC running while bazel build is on-going and any tooling removed. This is not a problem for something like `stack --nix` which uses a nix-shell which prevents the targets from being destroyed. We do not have that luxury with bazel.

However having said that, maybe we don't actually care. nix GC is no different than user removing something from `/usr/bin` through their package manager for example. We can take the same stance.

On the other hand, bazel itself does things a little bit differently: the tools that it has will usually be downloaded and stored by it: that makes them "guaranteed" to be present modulo malicious intent. The fundamental difference is that one is managed by bazel and the other by nix.

If we can't find a good, reliable solution that lets us keep targets that are in-use (for example by creating GC roots to everything in bazel's `external/`: this is possible but likely not viable to automate so that it's invisible from the user), we should at least document this drawback.

mrkkrp:
So they are not garbage collected.

Closes #7.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -60,7 +60,7 @@ def _nixpkgs_package_impl(ctx):
   if nix_build_path == None:
     fail("Could not find nix-build on the path. Please install it. See: https://nixos.org/nix/")
 
-  nix_build = [nix_build_path] + path + ["--no-out-link"] + expr_args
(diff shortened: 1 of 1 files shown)


mboes:
Closes #7.

PR #16, opened 2018-06-08
Allow to re-configure cc toolchain using Nix

mrkkrp:
A solution to tweag/rules_haskell#250.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1,3 +1,5 @@
+load("@bazel_tools//tools/cpp:cc_configure.bzl", "cc_autoconf_impl")
+
 """Rules for importing Nixpkgs packages."""
 
 def _nixpkgs_git_repository_impl(ctx):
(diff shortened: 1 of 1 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -101,3 +103,68 @@ nixpkgs_package = repository_rule(
...
+    "ld": attr.label(
+      executable=True,
+      cfg="host",
+      allow_single_file=True,
+      doc="`ld` to use in cc toolchain",
+    ),
+  },
+  local = True,
+  environ = [
+        "ABI_LIBC_VERSION",
```
mboes: Where does this list of environment variables come from?

mrkkrp: The original rule [here](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl#L58) depends on these env vars. Since we're re-using implementation here, it makes sense that our rule also necessarily depends on the same set of variables.

nixpkgs/nixpkgs.bzl
```diff
@@ -101,3 +103,68 @@ nixpkgs_package = repository_rule(
...
+    "ld": ctx.path(ctx.attr.ld),
+  }
+  return cc_autoconf_impl(ctx, overriden_tools)
+
+cc_configure_custom = repository_rule(
+  implementation = _cc_configure_custom,
+  attrs = {
+    "gcc": attr.label(
+      executable=True,
+      cfg="host",
```
mboes: This line looks suspicious to me. Why is it needed?

nixpkgs/nixpkgs.bzl
mboes: Yes but if we make a nixpkgs specific rule, then by dropping *all* these env variables we make the build more hermetic, and avoid getting subtle bugs down the line because of a mismatch in the env vars we import vs those the upstream rule imports. Cross compilation is not supported by this rule in any case.

nixpkgs/nixpkgs.bzl
mrkkrp: If we want `executable` it is obligatory to specify `cfg="host"`, see: https://docs.bazel.build/versions/master/skylark/rules.html#configurations.

> For each label attribute, you can decide whether the dependency should be built in the same configuration, or transition to the host configuration (using cfg). If a label attribute has the flag executable=True, the configuration must be set explicitly.

And later:

> Tools that are executed as part of the build (e.g., compilers, code generators) should be built for the host configuration. In this case, specify cfg="host" in the attribute.

mboes: OK makes sense.

mboes:
Closing for now, since the current best practice is to run rules_haskell inside a nix-shell when a toolchain is sourced from Nixpkgs.

Issue #17, opened 2018-06-12
Incorrect cache behavior for `nixpkgs_package`


guibou:
`nixpkgs_package` caching is based on the parameters of the rule, such as `nix_file` or `nix_file_content`. They can reference other files which are not taken into account by the cache.

This mean that any change in these files does not invalidate the cache.

*Proposed new design* (I'll evaluate feasibility soon, bazel needs a way to force a rule to be run and control over the caching mecanism), the rule should always been build, but the caching / invalidation should be based on the result of the underlying `nix-build` command.

Fuuzetsu:
I see the problem but I don't know the solution. I suspect that bazel should be running the command in some sort of sandbox where external files are not available unless they are part of the rule input at which point they are cached as usual. This is the only sane way I can think of.

>rule should always been build, but the caching / invalidation should be based on the result of the underlying nix-build command

https://github.com/bazelbuild/bazel/issues/3041

Fuuzetsu:
side-note: I don't have ability to label in this repo, can I get that back? Thanks.

guibou:
> bazelbuild/bazel#3041

There is no much activity in this area since a few, so I guess that we can conclude that `bazel` won't provide a solution in a short delay.

A pragmatic solution can be to add a `data` attribute to `nipkgs_rules` which list all the transitive file dependencies. It will be error prone, but at least we'll get a correct cache behavior if this attribute is correctly setup.

Fuuzetsu:
I wouldn't be against such a well-documented workaround.

facundominguez:
Is it possible to have a rule modify one of its inputs? Maybe if it is not sandboxed? If so, it could be used to trigger any other rule that depends on it every time that bazel is executed.

mboes:
#18 only partially solves this. There is another issue orthogonal to what #18 resolves but related to this ticket. If `<nixpkgs>` changes, say following a `nix-channel --update`, then the cache won't be invalidated. Currently, we implicitly use `<nixpkgs>` if no `repository` attribute was provided. If we remove that feature, and force the user to always name a repository by git tag or commit, then we avoid that problem.

mboes:
Note that `nix-build`, even when it's a noop, still costs a fair amount of time. So we should avoid running it unconditionally.

guibou:
@mboes I'll do a second PR to remove the implicit `<nixpkgs>`.

[...]

PR #18, opened 2018-06-13
Introduce a `data` attribute to `nixpkgs_package`

guibou:
This is a solution for #17, the new `data` attribute references all the files
needed to execute the nix expression, hence invalidating cache if they are
modified.


guibou:
@mboes the point of using `data` was that the "root" nix file was specified in `nix_file` and all the dependencies are in `data`. I can rename it `srcs` if you wish.

But what about `nix_file`, should we keep it or delete it? If we remove it, which file in `srcs` should be used? The order is not reliable because user may use `glob` to fill `srcs`. Should we use `srcs` AND `nix_file`?

Fuuzetsu:
@mboes I don't think the issue is only up to `nix` files and data was a better name. The nix file itself can refer to arbitrary files: if the arbitrary file changes and the nix file doesn't, currently bazel will not re-run the role. I thought this was the original problem. Of course these arbitrary files could be other nix files but that's only a subset of possible inputs.

guibou:
@mboes name changed from `data` to `srcs`. And one of `repository` or `path` attribute should be setup.

mboes:
@Fuuzetsu `data` is typically used for runtime dependencies. But there is no "runtime" here, and these are dependencies of the package metadata description. Technically the `build_file` could also have dependencies, since it can `load()`. So opting for renaming to `nix_file_deps`.

PR #25, opened 2018-08-08
Force explicit nixpkgs

### LINKED ISSUE #24: nixpkgs_package shouldn't default to <nixpkgs>
When calling `nixpkgs_package` without any of the `path` or `repository` argument, `bazel` complains:

```
DEBUG: /home/guillaume/.cache/bazel/_bazel_guillaume/9923957c70cd7b59bec21e32af97c8be/external/io_tweag_rules_nixpkgs/nixpkgs/nixpkgs.bzl:66:5: 
WARNING: Implicitly using '<nixpkgs>' as the location of Nixpkgs.
This is not recommended because it makes the build non-hermetic.
Set which Nixpkgs to use explicitly using 'repository' or 'path' attributes.
```

My setup is correct and hermetic because the file pointed by `nix_file` setups its own pinned clone of `nixpkgs`. To remove the warning, I have to set a dummy `path = "."`.

I propose the following changes:

- remove the defaulting to `<nixpkgs>` when neither `path` nor repository are selected. and instead load the provided `nix_file` (or `nix_file_content`) with no value for `<nixpkgs>`.
- (optional) add a new attribute `enable_impure_nixpkgs_source`, with default to `false`, which re-enable the current behavior.

This way, by default, the user *MUST* provides either `path` or `repository` or a `nix_file` with a pinned `nixpkgs`. The optional new attribute may allow her to get back the unsafe behavior.

I'm opening the issue for discussion, PR will follow if we agree on the design.

This may be a good final conclusion to #17

guibou:
This closes #24 by forcing an explicit `nixpkgs` clone.

- If the user provides no `nix_file`, and no `path` or `repository`, it fails with a bazel error
- If the user provides a `nix_file`, there is no solution to detect if the file is indeed using a pinned nixpkgs clone. We just override `NIX_PATH` to an invalid path and hop for the best

I'm not happy with the way I'm abusing `NIX_PATH` to write an error message. Any better idea?

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -30,6 +30,10 @@ def _nixpkgs_package_impl(ctx):
   else:
     ctx.template("BUILD", Label("@io_tweag_rules_nixpkgs//nixpkgs:BUILD.pkg"))
 
+  strFailureImplicitNixpkgs = (
+     "One of 'path', 'repository', 'nix_file' or 'nix_file_content' must be provided. "
+     + "The NIX_PATH environment variable is not inherited.")
+
   expr_args = []
   if ctx.attr.nix_file and ctx.attr.nix_file_content:
     fail("Specify one of 'nix_file' or 'nix_file_content', but not both.")
@@ -37,6 +41,8 @@ def _nixpkgs_package_impl(ctx):
     ctx.symlink(ctx.attr.nix_file, "default.nix")
   elif ctx.attr.nix_file_content:
     expr_args = ["-E", ctx.attr.nix_file_content]
+  elif not (ctx.attr.path or ctx.attr.repository):
+    fail(strFailureImplicitNixpkgs)
   else:
     expr_args = ["-E", "import <nixpkgs> {}"]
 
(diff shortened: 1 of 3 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -82,7 +88,17 @@ Set which Nixpkgs to use explicitly using 'repository' or 'path' attributes.
...
-  res = ctx.execute(nix_build, quiet = False, timeout = timeout)
+
+  # Here we override NIX_PATH to something silly
+  # hence it will be impossible to use a implicit nixpkgs repository
+
+  # XXX: we have no way to ensure that the included 'nix_file' or
+  # 'nix_file_content' point to a pinned nixpkgs repository, so
+  # 'NIX_PATH' contains an error message, which will be seen by bazel user
+  res = ctx.execute(nix_build, quiet = False, timeout = timeout, environment=dict(
+    NIX_PATH="nixpkgs=implicit nixpkgs value is disabled for hermeticity reason."
```
mboes: Why do this? Setting `NIX_PATH=` (i.e. to the empty string) works equally well.

nixpkgs/nixpkgs.bzl
```diff
@@ -30,13 +30,19 @@ def _nixpkgs_package_impl(ctx):
   else:
     ctx.template("BUILD", Label("@io_tweag_rules_nixpkgs//nixpkgs:BUILD.pkg"))
 
+  strFailureImplicitNixpkgs = (
+    "You must provide a 'path' or 'repository' or a nix file which point to a"
```
mboes: Simplify the error message:

> One of 'path' or 'repository' must be provided. The NIX_PATH environment variable is not inherited.

nixpkgs/nixpkgs.bzl
```diff
@@ -65,12 +71,12 @@ def _nixpkgs_package_impl(ctx):
     path = ["-I", "nixpkgs={0}".format(ctx.path(ctx.attr.repository).dirname)]
   elif ctx.attr.path:
     path = ["-I", "nixpkgs={0}".format(ctx.attr.path)]
+  elif not ctx.attr.nix_file and not ctx.attr.nix_file_content:
```
mboes: Use De Morgan's law for a more readable guard.

nixpkgs/nixpkgs.bzl
mboes: This is like using 0 as an error value instead of a Maybe type, or Verisign replacing all "Domain not found" errors with [Site Finder](https://en.wikipedia.org/wiki/Site_Finder). a) we should not be in the business of abusing environment variables meant for structured *input* to the Nix interpreter as unstructured *output* to the user. b) we should not assume that users can't follow a structured debugging process. Give them good documentation and get out of their way.

guibou: Ok.

nixpkgs/nixpkgs.bzl
mboes: Then adjust the error message accordingly.

> One of 'path', 'repository', `nix_file` or `nix_file_content` must be provided. The NIX_PATH environment variable is not inherited.

Error messages should be pithy, to the point and assume as little about what the user is doing as possible.

guibou:
@mboes  I addressed the point of your reviews, please have a new look.

nixpkgs/nixpkgs.bzl
guibou: I agree that the solution is hackish, but I rather prefer that than an error with limited details.

nixpkgs/nixpkgs.bzl
guibou: I disagree about readability, but I'm sure that's a matter of personal taste, I'll follow yours for that one.

guibou:
@mboes I addressed your review, added a `CHANGELOG` entry and updated the documentation.

PR #27, opened 2018-08-29
Improve symlink and refactor

Profpatsch:
This refactors a few things and removes a two bugs with symlinking.

I split it into a four consecutive commits with one standalone change each, that should be helpful for reviewing.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -57,7 +57,13 @@ def _nixpkgs_package_impl(ctx):
     "-A", ctx.attr.attribute_path
           if ctx.attr.nix_file or ctx.attr.nix_file_content
           else ctx.attr.attribute_path or ctx.attr.name,
-    "--no-out-link",
+    # Creating an out link prevents nix from garbage collecting the store path.
+    # nixpkgs uses `nix-support/` for such house-keeping files, so we mirror them
+    # and use `bazel-support/`, under the assumption that no nix package has
+    # a file named `bazel-support` in its root.
+    # A `bazel clean` deletes the symlink and thus nix is free to garbage collect
+    # the store path.
+    "--out-link", "bazel-support/nix-out-link"
   ])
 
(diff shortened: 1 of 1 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -99,11 +99,9 @@ def _nixpkgs_package_impl(ctx):
   # Build a forest of symlinks (like new_local_package() does) to the
   # Nix store.
 
-  find_path = ctx.which("find")
-  if find_path == None:
-    fail("Could not find the 'find' command. Please ensure it is in your PATH.")
+  find_path = _executable_path("find", ctx)
 
-  res = ctx.execute(["find", output_path, "-maxdepth", "1"])
+  res = ctx.execute([find_path, output_path, "-maxdepth", "1"])
```
Profpatsch: This actually fixes a bug in some situations, `find_path` should have been used here all along.

nixpkgs/nixpkgs.bzl
```diff
@@ -98,11 +99,9 @@ def _nixpkgs_package_impl(ctx):
...
```
Profpatsch: This actually fixes a bug in some situations, `find_path` should have been used here all along.

Profpatsch:
> I believe we should create an out link to guard against that, though an out-link from within the bazel cache.

That’s a great idea. I think we need to create the out link in the build folder, which might clash with derivation files in the worst case. Nix already has the problem and introduces a canonical path called `nix-support` where it puts such house-keeping files. So I think it’s best to mirror that and create the out link in `bazel-support/nix-out-link`.

Profpatsch:
Added support in a new commit. An example: 

```
$ readlink /home/philip/.cache/bazel/_bazel_philip/aec3516d7ea5cc1a3eaf026b7b4beb22/external/go/bazel-support/nix-out-link
/nix/store/g85kj9nnmjwiv602n1m2igyi8ckh6y43-go-1.10.3
```

PR #29, opened 2018-09-19
Allow passing several labels to `nixpkgs_package`

regnat:
Deprecates the `repository` attribute in favor of the more general
`repositories` which expects a dict of the form

```python
{
  "@some_label": "some_name",
  "@some_other_label": "some_other_name",
}
```

and generating the `NIX_PATH`
`some_name=path(@some_label):some_other_name=path(@some_other_label)`

-----------

I also have a couple of questions regarding this:

- I would prefer having the dict be `{ "name": "@label"}` instead of `{ "@labelb: "name"}`, but the docs at https://docs.bazel.build/versions/master/skylark/lib/attr.html only mentions `label_keyed_string_dict` and not something like `string_keyed_label_dict`. Does anyone knows whether there is a way to build custom types for attrs?
- I have manually updated the documentation in the `README`. Is it the right way to do so or is it generated in some way?

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -21,6 +21,15 @@ nixpkgs_git_repository = repository_rule(
 )
 
 def _nixpkgs_package_impl(ctx):
+  repositories = None
+  if ctx.attr.repositories:
+    repositories = ctx.attr.repositories
+
+  if ctx.attr.repository:
+    print("The 'repository' attribute is deprecated, use 'repositories' instead")
+    repositories = { ctx.attr.repository: "nixpkgs" } + \
+        (repositories if repositories else {})
+
   if ctx.attr.build_file and ctx.attr.build_file_content:
     fail("Specify one of 'build_file' or 'build_file_content', but not both.")
   elif ctx.attr.build_file:
@@ -31,7 +40,7 @@ def _nixpkgs_package_impl(ctx):
     ctx.template("BUILD", Label("@io_tweag_rules_nixpkgs//nixpkgs:BUILD.pkg"))
 
   strFailureImplicitNixpkgs = (
-     "One of 'path', 'repository', 'nix_file' or 'nix_file_content' must be provided. "
+     "One of 'path', 'repositories', 'nix_file' or 'nix_file_content' must be provided. "
      + "The NIX_PATH environment variable is not inherited.")
 
   expr_args = []
@@ -41,7 +50,7 @@ def _nixpkgs_package_impl(ctx):
     ctx.symlink(ctx.attr.nix_file, "default.nix")
   elif ctx.attr.nix_file_content:
     expr_args = ["-E", ctx.attr.nix_file_content]
-  elif not (ctx.attr.path or ctx.attr.repository):
+  elif not (ctx.attr.path or repositories):
     fail(strFailureImplicitNixpkgs)
   else:
     expr_args = ["-E", "import <nixpkgs> {}"]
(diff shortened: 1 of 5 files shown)


regnat:
Mh I've absolutely no idea what the CI failure means and it doesn't seem related to my work…

mboes:
Looks like an upstream regression: https://github.com/bazelbuild/bazel/pull/5946. We're hit by this because we're not currently pinning a nixpkgs version in this repository.

nixpkgs/nixpkgs.bzl
```diff
@@ -106,21 +107,46 @@ def _nixpkgs_package_impl(ctx):
...
+          # no nix package has a file named `bazel-support` in its root.  A
+          # `bazel clean` deletes the symlink and thus nix is free to garbage
+          # collect the store path.
+          "--out-link", "bazel-support/nix-out-link"
+        ],
+        *args,
+        **kwargs
+    )
+
+def nix_instantiate(*args, **kwargs):
```
mboes: Is this function used anywhere? It looks like dead code to me.

nixpkgs/nixpkgs.bzl
```diff
@@ -106,21 +107,46 @@ def _nixpkgs_package_impl(ctx):
...
     "path": attr.string(),
+    "repositories": attr.label_keyed_string_dict(),
     "repository": attr.label(),
     "build_file": attr.label(),
     "build_file_content": attr.string(),
   },
   local = True,
 )
 
+def nixpkgs_package(*args, **kwargs):
```
mboes: What's the reason for changing this rule to a macro that calls a single rule?

nixpkgs/nixpkgs.bzl
regnat: Ouch sorry, pushed the wrong commit :/

Should be fixed now

mboes:
See https://github.com/bazelbuild/bazel/issues/5356 for the upstream ticket.

regnat:
> Could you say more about your use case?

It's motivated by tweag/rules_haskell#442 where I need `rules_haskell` to export a nix function (in `haskell/nix/default.nix`) which has to be imported and used in the dependent repositories.
The logical solution for that is to add it to the  `NIX_PATH`, which requires this API extension (but obviously if someone has a better solution to this, I'm all for it)

mboes:
That's a good idea. Provided it solves the forward compatibility problem. That's the main constraint.

Profpatsch:
> The logical solution for that is to add it to the `NIX_PATH`

I think the idea behind the change is sensible, but I don’t think you necessarily need to add it to `NIX_PATH`. It’s just a convenience to fill free variables of the form `<foo>` and can always be replaced by simple lambdas (as long as the path is a local file). That is you can change

```
env NIX_PATH='foo=./foo:bar=./bar' nix-instantiate -E "<foo> <bar>"
```

to

```
nix-instantiate -E "{foo, bar}: foo bar" --arg foo "./foo" --arg bar "./bar"
```

regnat:
@Profpatsch that's true (and being able to pass arguments to the nix expression would be cool), but the `NIX_PATH` case is simpler as you know it will only contain paths, so it makes sense to put only labels in it while the arguments can be anything and since labels are fairly magical you can't say something like "this field can be a string or a label".

regnat:
I just stumbled upon another use-case for this: We have a master project using `rules_nixpkgs` and a custom (in tree) nix overlay and we want to create new projects depending on this master project. These child projects will use `rules_nixpkgs` as well and their nix config should include the overlay of the master.

Afaik there is no easy way to do this at the moment, while with this change it can be done by adding `{ "master_overlay": "@master_project//:overlay" }` to the `repositories` field of each `nixpkgs_package` statement (and referencing it as `<master_overlay>` in the nix file)

nixpkgs/nixpkgs.bzl
```diff
@@ -94,7 +104,7 @@ def _nixpkgs_package_impl(ctx):
   timeout = 1073741824
 
   res = ctx.execute(nix_build, quiet = False, timeout = timeout,
-                    environment=dict(NIX_PATH="nixpkgs=" + nix_path))
+                    environment=dict(NIX_PATH=nix_path))
```
Profpatsch: This breaks the `path` argument, since it’s not of the form `"nixpkgs=path"` anymore.

Profpatsch:
I’d rename `repositories` to `nix_path`, to follow nix nomenclature more closely, and deprecate `path` in the process.

Profpatsch:
Rebased the changes.

PR #31, opened 2018-09-20
Pin nixpkgs revision in the CI

regnat:
Prevent failures such as
https://circleci.com/gh/tweag/rules_nixpkgs/62?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
due to upstream changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,8 +10,9 @@ jobs:
       - run:
           name: System dependencies
           command: |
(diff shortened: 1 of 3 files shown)


Profpatsch:
Shouldn’t we used the trick from yesterday and create a `/tests/nixpkgs.nix` with `import (fetchTarball …)` and then set `NIX_PATH` to that file?

regnat:
That sounds right indeed, I'll do that

PR #35, opened 2018-10-24
Improve documentation of `build_file` and add tests

### LINKED ISSUE #9: Consider exposing all files in a filegroup
Currently we only expose `.so` files from `lib` and things from `bin`: this is quite restrictive: we can't use header files or use binaries that don't live directly in `bin` &c.

Is there any good reason to not have an `out` file group that user can reach for?

Related question: can use filter filegroups? That'd be very useful.

Profpatsch:
The `BUILD` file in the repository created by `nixpkgs_package` had
three filegroups that exposed `bin`, `include` and `lib`, but none
that exposed all files of the nix build output.
We introduce a new target named `:output` that does just that.

fixes #9


Profpatsch:
Point taken, I’ve instead opted to improve the user documentation for `build_file` and add two tests.

Profpatsch:
CircleCI is red, #36 needs to be merged before it prints the error logs so I can debug this.

Fuuzetsu:
I guess I'm happy with #9 being closed because it's documented how to access the other files.

Profpatsch:
Nice, had some bashisms in the `test_output` script, but green now.

PR #40, opened 2018-11-10
Add nixpkgs_cc_configure()

mboes:
A call to this macro in your WORKSPACE file is a simple one-liner. It
declares a nixpkgs_package with all the tools we need for a CC
toolchain, and then gets Bazel to use them. Making sure the Bazel uses
the C compiler from Nixpkgs rather than whatever it autodetects on the
PATH improves build hermeticity, and side steps a whole host of issues
we've been facing on rules_haskell due to system GCC wanting to link
against the system GLIBC while Nixpkgs GCC wants another one.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -97,14 +97,17 @@ def _nixpkgs_package_impl(repository_ctx):
   # as we can do.
   timeout = 1073741824
 
-  res = repository_ctx.execute(nix_build, quiet = False, timeout = timeout,
-                    environment=dict(NIX_PATH=nix_path))
-  if res.return_code == 0:
-    output_path = res.stdout.splitlines()[-1]
-  else:
-    _execute_error(res, "Cannot build Nix attribute `{}`"
-                          .format(repository_ctx.attr.attribute_path))
-
+  exec_result = _execute_or_fail(
+    repository_ctx,
+    nix_build,
+    failure_message = "Cannot build Nix attribute '{}'.".format(
+      repository_ctx.attr.attribute_path
+    ),
+    quiet = False,
+    timeout = timeout,
+    environment=dict(NIX_PATH=nix_path),
+  )
+  output_path = exec_result.stdout.splitlines()[-1]
   # Build a forest of symlinks (like new_local_package() does) to the
   # Nix store.
   _symlink_children(repository_ctx, output_path)
@@ -142,6 +145,26 @@ def nixpkgs_package(*args, **kwargs):
   else:
     _nixpkgs_package(*args, **kwargs)
 
+def _execute_or_fail(repository_ctx, arguments, failure_message = "", *args, **kwargs):
+  """Call repository_ctx.execute() and fail if non-zero return code."""
+  result = repository_ctx.execute(arguments, *args, **kwargs)
+  if result.return_code:
+    outputs = dict(
+      failure_message = failure_message,
+      arguments = arguments,
+      return_code = result.return_code,
+      stderr = result.stderr,
+    )
+    fail("""
+{failure_message}
+Command: {arguments}
+Return code: {return_code}
+Error output:
+{stderr}
+""").format(**outputs)
+  return result
+
+
 def _symlink_children(repository_ctx, target_dir):
   """Create a symlink to all children of `target_dir` in the current
   build directory."""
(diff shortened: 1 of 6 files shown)


mboes:
This is #16 reopened.

nixpkgs/nixpkgs.bzl
```diff
@@ -185,3 +186,63 @@ stderr:
...
+
+  # Make a list of all available tools in the Nix derivation. Override
+  # the Bazel autoconfiguration with the tools we found.
+  res = repository_ctx.execute(["ls", workspace_root + "/bin"])
+  if res.return_code == 0:
+    overriden_tools = {
+      tool: repository_ctx.path(Label("@nixpkgs_cc_toolchain//:bin/" + tool))
+      for tool in res.stdout.splitlines()
+    }
+    cc_autoconf_impl(repository_ctx, overriden_tools = overriden_tools)
```
Profpatsch: The rule uses `repository_ctx.which` in `unix_cc_configure.bzl` to resolve paths passed through `overridden_tools`, and it’s not clear from the documentation what that means for the absolute paths we pass it: https://github.com/bazelbuild/bazel/issues/6660 (I’m too lazy to look into the Java source right now, but I assume it returns absolute paths unmodified).

nixpkgs/nixpkgs.bzl
```diff
@@ -185,3 +186,63 @@ stderr:
...
+      }
+    """):
+  """Use a CC toolchain from Nixpkgs.
+  
+  By default, Bazel auto-configures a CC toolchain from commands (e.g.
+  `gcc`) available in the environment. To make builds more hermetic, use
+  this rule to specific explicitly which commands the toolchain should
+  use.
+  """
+  if repository and repositories or not repository and not repositories:
```
Profpatsch: I suspect `and` binds stronger than `or`? I’d add parens to make it obvious.

nixpkgs/nixpkgs.bzl
```diff
@@ -185,3 +186,63 @@ stderr:
...
+    repositories = {"nixpkgs": repository}
+
+  nixpkgs_package(
+    name = "nixpkgs_cc_toolchain",
+    repositories = repositories,
+    nix_file_content = nix_file_content,
+    build_file_content = """exports_files(glob(["bin/*"]))""",
+  )
+  _nixpkgs_cc_autoconf(name = "local_config_cc")
+  native.bind(name = "cc_toolchain", actual = "@local_config_cc//:toolchain")
```
Profpatsch: [Using `bind` is not recommended](https://docs.bazel.build/versions/master/be/workspace.html#bind). Afaik it’s not deprecated though.

mboes: This code was blindly copied from `cc_autoconf`. I think we should do whatever they are doing.

nixpkgs/nixpkgs.bzl
mboes: Always does. It's a convention from mathematics akin to `*` always stronger than `+`.

nixpkgs/nixpkgs.bzl
```diff
@@ -185,3 +186,63 @@ stderr:
...
+    ["dirname", workspace_file_path]
+  ).stdout.rstrip()
+
+  # Make a list of all available tools in the Nix derivation. Override
+  # the Bazel autoconfiguration with the tools we found.
+  res = repository_ctx.execute(["ls", workspace_root + "/bin"])
+  if res.return_code == 0:
+    overriden_tools = {
+      tool: repository_ctx.path(Label("@nixpkgs_cc_toolchain//:bin/" + tool))
+      for tool in res.stdout.splitlines()
```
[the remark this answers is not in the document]
mboes: Indent to where? This is a comprehension. I usually see it indented this way.

nixpkgs/nixpkgs.bzl
```diff
@@ -185,3 +186,63 @@ stderr:
...
+  workspace_file_path = repository_ctx.path(
+    Label("@nixpkgs_cc_toolchain//:WORKSPACE")
+  )
+  workspace_root = repository_ctx.execute(
+    ["dirname", workspace_file_path]
+  ).stdout.rstrip()
+
+  # Make a list of all available tools in the Nix derivation. Override
+  # the Bazel autoconfiguration with the tools we found.
+  res = repository_ctx.execute(["ls", workspace_root + "/bin"])
```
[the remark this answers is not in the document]
mboes: OK I split out a `_find_children()` function. And I'm calling that now instead of `ls`.

mboes:
@Profpatsch PTAL

nixpkgs/nixpkgs.bzl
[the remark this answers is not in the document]
Profpatsch: Looking good!

PR #42, opened 2018-11-13
Refactor: Replace _execute_error() with _execute_or_fail().

mboes:
This refactor has two simple goals:

* Simplify control flow: callers no longer need to `if-then-else`
  themselves, which becomes cumbersome when you have a sequence of
  calls to `repository_ctx.execute()`.
* Improve the output on failure (by listing the failing command).

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -97,14 +97,17 @@ def _nixpkgs_package_impl(repository_ctx):
   # as we can do.
   timeout = 1073741824
 
-  res = repository_ctx.execute(nix_build, quiet = False, timeout = timeout,
-                    environment=dict(NIX_PATH=nix_path))
-  if res.return_code == 0:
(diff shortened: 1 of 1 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -142,6 +145,26 @@ def nixpkgs_package(*args, **kwargs):
   else:
     _nixpkgs_package(*args, **kwargs)
 
+def _execute_or_fail(repository_ctx, arguments, failure_message = "", *args, **kwargs):
```
Profpatsch: Does `*args` work after `failure_message=`? I’m forgot most of the details of this pythonism.

mboes: Yes.

mboes:
0 _should_ always mean success. This convention is followed so universally that I don't think we need to pessimize our code readability just in case.

PR #45, opened 2018-11-15
Implement nixpkgs_local_repository()

mboes:
This rule does not add any new expressive power, because we can
already provide a Nix expression in a local file directly using the
`repository` or `repositories` attributes. The arguments for adding it
are:

* consistency. Where there is a native `local_repository` and
  `git_repository`, there are now rules_nixpkgs equivalents of both.
* Abstraction. When the Nixpkgs we want to use is pulled directly from
  Git, we don't have much of a choice but to declare a new external
  repository for it in Bazel. We can the refer to this repository
  simply as `@nixpkgs` in all `nixpkgs_package` declarations. It's
  better if whether the repository comes from Git or is local becomes
  an irrelevant concern at call sites, i.e. that it is abstracted
  away.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1,6 +1,8 @@
+"""Rules for importing Nixpkgs packages."""
+
+
 load("@bazel_tools//tools/cpp:cc_configure.bzl", "cc_autoconf_impl")
 
-"""Rules for importing Nixpkgs packages."""
 
 def _nixpkgs_git_repository_impl(repository_ctx):
   repository_ctx.file('BUILD')
@@ -13,6 +15,7 @@ def _nixpkgs_git_repository_impl(repository_ctx):
     sha256 = repository_ctx.attr.sha256,
   )
 
+
 nixpkgs_git_repository = repository_rule(
   implementation = _nixpkgs_git_repository_impl,
   attrs = {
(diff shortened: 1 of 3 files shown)


mboes:
> Am I right to assume that with this rule, forgetting to include some files in nix_file_deps will cause the build to fail instead of silently insert impurities ?

I'm not sure. Though whatever the case happens to be, the same applies to `nixpkgs_package`, because it's using a common code path to handle this particular attribute.

nixpkgs/nixpkgs.bzl
```diff
@@ -23,6 +26,38 @@ nixpkgs_git_repository = repository_rule(
...
+    repository_ctx.file(
+      path = "default.nix",
+      content = repository_ctx.attr.nix_file_content,
+      executable = False,
+    )
+    target = repository_ctx.path("default.nix")
+  else:
+    target = repository_ctx.path(repository_ctx.attr.nix_file)
+    repository_ctx.symlink(target, target.basename)
+  repository_ctx.symlink(target.basename, repository_ctx.name)
```
Profpatsch: This is the symlink that produces the file for the `@foo//:foo` target trick, right? I’d add an explanatory comment.

mboes: OK. I will duplicate the comment from the function just above.

PR #47, opened 2018-11-20
Make nixpkgs_cc_configure() a no-op on anything but Linux and Darwin.

mboes:
These are the two platforms supported by Nixpkgs currently.

cc @majcherm-da

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -214,6 +214,40 @@ def nixpkgs_cc_autoconf_impl(repository_ctx):
 
 nixpkgs_cc_autoconf = repository_rule(
     implementation = nixpkgs_cc_autoconf_impl,
+    # Copied from
+    # https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl.
+    # Keep in sync.
+    environ = [
+        "ABI_LIBC_VERSION",
+        "ABI_VERSION",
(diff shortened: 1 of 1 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -183,6 +184,11 @@ def nixpkgs_package(*args, **kwargs):
         _nixpkgs_package(*args, **kwargs)
 
 def _nixpkgs_cc_autoconf_impl(repository_ctx):
+    cpu_value = get_cpu_value(repository_ctx)
+    if cpu_value != "k8" or cpu_value != "darwin":
+        cc_autoconf_impl(repository_ctx)
+        return
```
majcherm-da: It would be good to pass `environ` list from: https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl#L59 into `nixpkgs_cc_autoconf` rule: https://github.com/tweag/rules_nixpkgs/blob/e9187aee2402febd375b4032c80fb524fed309d3/nixpkgs/nixpkgs.bzl#L216 so that it's aware of changes of env. variables which might affect the rule.

mboes: Urgh. You are right. In the case where we do use Nixpkgs tools none of those environment variables matter, so best to cull them. But in the Windows case, they might. So now we need to duplicate that long list of env vars... :cry:

Issue #49, opened 2018-11-27
Add a `nixpkgs_packageset` rule


thufschmitt:
Since https://github.com/tweag/rules_haskell/pull/442, the evaluation of the workspace of a rules_haskell-enabled repository takes an insane amount of time because it requires `nix-build`-ing every transitive haskell dependency, which easily means more than 200 `nix-build` calls.
We could speed things a lot by first `nix-instantiate`-ing a big json file mapping each haskell package to its `.drv`, and then just realize as needed these pre-intstantiated derivations. This would avoid the cost of evaluating nixpkgs 200 times (which is by far the bottleneck when everything is already built).

This trick can probably be used for more than just the `haskellPackages` case, so I propose we add a `nixpkgs_packages_set` rule to `rules_nixpkgs` which would be used like:

```python
nixpkgs_packages_set(
  name = "haskellPackages",
  repository = "@nixpkgs",
  base_attribute_path = "haskellPackages",
)
```

This would generate `@haskellPackages-base`, `@haskellPackages-streaming`,
`@haskellPackages-foobar`, … (and possibly one `@haskellPackages` with aliases
in it for all of these.

Internally this rule would

1. `nix-instantiate` a json file of the form

    ```json
    {
      "base": "/nix/store/…-base.drv",
      "streaming": "/nix/store/…-streaming.drv",
      "foobar": "/nix/store/…-foobar.drv",
      …
    }
    ```

2. Load this into a starlark dict one way or another

3. For each element of this dict, generate a call to a rule which would `nix-store --realize` the corresponding derivation

All this would require extending a bit the scope of nixpkgs (or doing thing in an ad-hoc way, not sure what's the best choice), to add

- A `nix_instantiate` rule allowing to import the result of `nix-instantiate`
- A `nix_realize` rule which would wrap `nix-store --realize`

Thoughs?

mboes:
I think this is a good idea. Though to be useful the json file you mention should be checked into the source, right? So that even the initial checkout isn't too slow? @zimbatm pointed out to me yesterday that this would be similar to Yarn lock files. And it so happens that @aehlig proposed to do something similar [here](https://blog.bazel.build/2018/09/28/first-class-resolved-file.html) and [here](https://github.com/bazelbuild/proposals/blob/master/designs/2018-11-07-design-recursive-workspaces.md).

regnat:
I didn't think of checking it. I think it would be a benefit even without doing it.

[...]

PR #50, opened 2018-12-04
WIP: nixpkgs_packages

regnat:
A first POC for a solution to #49.

This introduces a new `nixpkgs_packages` rule such that

```bzl
nixpkgs_packages(
    name = "foobarbaz",
    repositories = {"nixpkgs": "@nixpkgs"},
    packages = {
        "gcc": "gcc",
        "glibc": "glibc",
        "firefox": "firefox",
        "ghc": "ghc",
        },
)
```

will behave as if we had several individual `nixpkgs_package` rules (one for `gcc`, one for `glibc`, …) except that the instantiation of nixpkgs will be done only once instead of once per target.

This is a mix between my original proposal and the various suggestions.
Under the hood it will `nix-instantiate` a json file (with the rule ` nixpkgs_packages_instantiate`) of the form

```
{
  "firefox": "/nix/store/gcn62sk0qfg7w14kvxsv039lxli9sp0n-firefox-61.0.1.drv",
  "gcc": "/nix/store/vgl5acfn8calfavspngdbp68llqx408g-gcc-wrapper-7.3.0.drv",
  "ghc": "/nix/store/yjh0ihvhwlsby8pgwqn3lrl59319nf4v-ghc-8.4.3.drv",
  "glibc": "/nix/store/xazj2w1xc11s09ap5rgvki1znwj54dx2-glibc-2.27.drv"
}
```

and building each target will simply correspond to realizing the associated `drv` file (with the rule `nixpkgs_package_realize `).

This is only a POC right now, there are several points which would need to be sorted out:

- [ ] We might want to extend the interface a bit. In particular only a subset of the arguments of `nixpkgs_package` are supported − in particular `nix_file` and `nix_content` are not because this would require some more nix-side magic.
- [ ] There is a lot of duplicated logic between these new rules and `nixpkgs_package`. We could factor this out with a tiny bit of work, but if we extand the interface of the `nixpkgs_packages` macro enough, we could remove the `nixpkgs_package` rule altogether and replace it by a thin wrapper on top of `nixpkgs_packages`. However, this would make this simple rule more complicated.
- [ ] Right now I have to hack a bit because `nix-store --realize` doesn't play well with multiple-outputs derivations (see [this](https://github.com/tweag/rules_nixpkgs/blob/17748bb14d82c11682369a7fac1b2f2e348f308d/nixpkgs/drv_set_builder.nix#L6-L14)). If possible I'd rather get rid of it before we merge it.

/cc @zimbatm @Profpatsch @mboes


nixpkgs/nixpkgs.bzl
```diff
@@ -175,14 +178,169 @@ def nixpkgs_package(*args, **kwargs):
...
+
+    packages_record_inside = ";".join(
+        [
+            "\"{name}\" = nixpkgs.{attrPath}".format(name = name, attrPath = attrPath)
+            for (name, attrPath) in repository_ctx.attr.packages.items()
+        ]
+    )
+    packages_record = "nixpkgs: { " + packages_record_inside + "; }"
+
+    repository_ctx.file("packages_attributes_mappings.nix", packages_record)
```
zimbatm: is this path unique enough that two nixpkgs_packages can be defined in the same project?

nixpkgs/nixpkgs.bzl
```diff
@@ -175,14 +178,169 @@ def nixpkgs_package(*args, **kwargs):
...
+
+    nix_set_builder = repository_ctx.template("drv_set_builder.nix", Label("@io_tweag_rules_nixpkgs//nixpkgs:drv_set_builder.nix"))
+    nix_instantiate_args = [
+        "--eval",
+        "--strict",
+        "--read-write-mode",
+        "--json",
+        "drv_set_builder.nix",
+        "--arg", "packages", "import ./packages_attributes_mappings.nix",
```
zimbatm: in theory it should be possible to pass the package list straight as nix code in the argument and avoid the intermediary file. It also has the advantage that if the command it printed, all the context is available to run it manually.

nixpkgs/nixpkgs.bzl
regnat: Yes, that path is namespaced by the value of the `name` attribute of the rule which has to be unique

nixpkgs/nixpkgs.bzl
regnat: That's what I tried initially, but the original motivation for all that is to run it with the whole of `haskellPackages` which gave me `command-line too long` errors

regnat:
Closing until I or someone else manage to finish this

PR #51, opened 2018-12-04
Allow passing arbitrary arguments to nix-build

regnat:
Prerequisite for a proper solution to https://github.com/tweag/rules_haskell/pull/495#issuecomment-443521575

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -75,6 +75,15 @@ nixpkgs_package(
 )
 
 nixpkgs_package(
+    name = "extra-args-test",
+    nix_file_content = """
+{ packagePath }: (import <nixpkgs> {}).${packagePath}
+    """,
+    repository = "@nixpkgs",
+    nix_build_extra_args = ["--argstr", "packagePath", "hello"],
+)
+
+nixpkgs_package(
     name = "output-filegroup-manual-test",
     build_file_content = """
 package(default_visibility = [ "//visibility:public" ])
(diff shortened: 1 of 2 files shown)


Profpatsch:
I don’t really like escape hatches like these, because they break abstraction. What’s the proper solution for https://github.com/tweag/rules_haskell/pull/495#issuecomment-443521575 that needs to inject arbitrary strings into the argument list?

regnat:
@Profpatsch The case at hand is that we want to pass `-j 1` to `nix-build` to avoid an over-parallelisation of the builds.

Agreed, we could do that by adding an extra-argument to the rule, but looking at the man-page, I see quite a lot of other use-cases (passing extra-arguments to the nix expression with `--arg`, passing extra nix options with `--option` − would be useful for example for a project with a custom binary cache, etc..), and I'm not sure adding an option for each of these would be worth it

Profpatsch:
Since `rules_nixpkgs` is always executed inside bazel, and we don’t want to overparallelize the builds, doesn’t that speak for adding `-j 1` by default to our nix invocations?

regnat:
Yeah, that was the discussion on https://github.com/tweag/rules_haskell/pull/501#discussion_r238582872. The answer seems to be "yes", though that's something which should be configurable (because you might import one big nixpkgs package in which case you want to be able to parallelize inside your nix-build).

So putting the default to `-j 1` will be the natural consequence of this PR ;)

Profpatsch:
> though that's something which should be configurable (because you might import one big nixpkgs package in which case you want to be able to parallelize inside your nix-build).

I don’t really see why it should be configurable; the user of a monorepository cannot know what will actually be built in practice (except in extremely limited use-cases), so they should not try to micro-optimize stuff like this.

PR #63, opened 2019-01-25
Generalize unsupported platforms

NinjaTrappeur:
We need to evaluate some rules_nixpkgs's repository_rules in a WORKSPACE
before selecting and using the non-nix toolchain. The repository_rule
should provide a way not to fail on linux/darwin systems not having
`nix-build` in their $PATH

Dependency of https://github.com/tweag/rules_haskell/issues/629 .

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -56,12 +56,21 @@ nixpkgs_local_repository = repository_rule(
     },
 )
 
+def _is_supported_platform(repository_ctx):
+    if repository_ctx.execute(["nix-build","--version"]).return_code == 0:
+      nix_in_path = True
+    else:
+      nix_in_path = False
+    is_windows = repository_ctx.os.name.startswith("windows")
+    return (not is_windows) and nix_in_path
+
+
 def _nixpkgs_package_impl(repository_ctx):
     repository = repository_ctx.attr.repository
     repositories = repository_ctx.attr.repositories
 
     # Is nix supported on this platform?
-    not_supported = repository_ctx.os.name.startswith("windows")
+    not_supported = not _is_supported_platform(repository_ctx)
     # Should we fail if Nix is not supported?
     fail_not_supported = repository_ctx.attr.fail_not_supported
 


nixpkgs/nixpkgs.bzl
```diff
@@ -56,12 +56,21 @@ nixpkgs_local_repository = repository_rule(
     },
 )
 
+def _is_supported_platform(repository_ctx):
+    if repository_ctx.execute(["nix-build","--version"]).return_code == 0:
+      nix_in_path = True
+    else:
+      nix_in_path = False
+    is_windows = repository_ctx.os.name.startswith("windows")
```
Profpatsch: Can we do the windows check before the shellout and shortcut if it’s windows?

NinjaTrappeur: The whole windows check has been removed in https://github.com/tweag/rules_nixpkgs/pull/64

NinjaTrappeur:
That's a really good point.

> Is this really necessary?

It is, without this, we cannot evaluate the `WORKSPACE` where this rule is used if nix is not on `$PATH`.

However, there may be a more clever way to accomplish that. I don't have an immediate solution here but I'll spend some brain cycles on it.

NinjaTrappeur:
I'm wondering if checking if there's something in `$NIX_PATH` alone would be a good indicator that we have nix-build on `$PATH`.

NinjaTrappeur:
Actually `which` could be a better candidate here.

```
~ » bench "nix-build --version"                                                               
benchmarking nix-build --version
time                 17.94 ms   (17.39 ms .. 18.58 ms)
                     0.995 R²   (0.991 R² .. 0.998 R²)
mean                 17.65 ms   (17.21 ms .. 17.97 ms)
std dev              892.3 μs   (645.3 μs .. 1.176 ms)
variance introduced by outliers: 20% (moderately inflated)

~ » bench "which nix-build"                                                                   
benchmarking which nix-build
time                 3.952 ms   (3.791 ms .. 4.102 ms)
                     0.991 R²   (0.986 R² .. 0.995 R²)
mean                 3.911 ms   (3.835 ms .. 3.979 ms)
std dev              230.0 μs   (193.3 μs .. 283.2 μs)
variance introduced by outliers: 36% (moderately inflated)
```

Profpatsch:
There’s also a builtin `which`: https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#which

PR #64, opened 2019-01-25
Making cc_configure no-op on unsupported platforms

NinjaTrappeur:
`nixpkgs_cc_configure` should be no-op on platforms that does not have
a nixpkgs-based setup in order to let them evaluate their WORKSPACE.

Also a dependency for https://github.com/tweag/rules_haskell/issues/629

I'm not sure this one is as innocuous as the previous one. Could this break some setups in the wild?


nmattia:
Better to ask @mboes

PR #67, opened 2019-01-29
Follow symlinks of tools in nixpkgs_cc_configure

dajmaki:
We noticed poor http cache hit rates, even on the same machine when running the
same build from a different working tree. By diffing the execution log it was clear that
the path to cc_toolchain was going via the workspace's output_path, which subsequently 
caused mismatching hashes.

This changes nixpkgs_cc_configure to follow symlinks of the binaries given to overriden_tools.
With this change the execution log shows nix-store paths as expected and the cache hit rates
are as expected.


mboes:
@shmish111 might be the cause of what you were noticing too.

PR #70, opened 2019-02-18
nixpkgs_packages

Profpatsch:
Re-open of https://github.com/tweag/rules_nixpkgs/pull/50, rebased onto current master.

Every change done by @regnat 

Fixes https://github.com/tweag/rules_nixpkgs/issues/49


WORKSPACE
```diff
@@ -83,16 +86,35 @@ nixpkgs_package(
...
+    packages = {
+        "nixpkgs-packages-hello": "hello",
+        "gcc": "gcc",
+        "glibc": "glibc",
+        "firefox": "firefox", # Won't be built until we explicitly ask for it
+        "ghc": "ghc",
+        "nixpkgs-packages-hello-from-file": { "nix_file": "//tests:hello.nix" },
+        "nixpkgs-packages-hello-from-expr": { "nix_file_content": "let pkgs = import <nixpkgs> {}; in pkgs.hello" },
+        "output-filegroup-manual-test": {
+            "build_file_content": """
```
Profpatsch: Should add an example where the evaluation fails.

nixpkgs/nixpkgs.bzl
```diff
@@ -59,125 +59,25 @@ nixpkgs_local_repository = repository_rule(
...
+    ):
+  nixpkgs_packages(
+      name = name + "__drvfile",
+      repository = repository,
+      repositories = repositories,
+      nixopts = nixopts,
+      packages = {
+          name: dict(
+              nixops = nixopts,
+              **kwargs
```
Profpatsch: We can’t fail with both `nix_file` and `attribute_path` are supplied, that is a regression.

Instead, it should ask for the attribute in `nix_file`.

nixpkgs/nixpkgs.bzl
```diff
@@ -374,9 +296,21 @@ nixpkgs_package_realize = repository_rule(
...
 )
 
+def hasAttr (item, attrName):
+    """
+    Checks whether the given item is a record with the given field set to
+    non-None
+    """
+    return type(item) == type({}) and \
+        attrName in item and \
+        item[attrName] != None
```
Profpatsch: You have some other code in an earlier commit that can use this function.

nixpkgs/nixpkgs.bzl
```diff
@@ -206,6 +206,104 @@ def invert_repositories(f, *args, **kwargs):
...
 
+def _nixpkgs_packages_instantiate_impl(repository_ctx):
+    repository_ctx.file("BUILD", "exports_files([\"nix_attrs.nix\"])")
+
+    nix_instantiate_path = _executable_path(
+        repository_ctx,
+        "nix-instantiate",
+        extra_msg = "See: https://nixos.org/nix/",
+    )
+
```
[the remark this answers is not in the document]
regnat: The only factorizable repetition here is the `"{foo} = {bar}"` template, and I think that putting that into an auxiliary function will be more cognitive overhead than repeating it twice (especially since functions must be defined at top-level so it would be defined relatively far away from the only place it's used)

nixpkgs/drv_set_builder.nix
```diff
@@ -0,0 +1,17 @@
+{ packages
+, nixpkgs ? import <nixpkgs> {}
```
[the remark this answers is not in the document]
regnat: It's the default 99.9% of the times this file will be called, so I don't think it's worth bothering with passing it explicitly (esp. since everyone expects most top-level nix file to import <nixpkgs> anyways)

nixpkgs/nixpkgs.bzl
```diff
@@ -206,6 +206,104 @@ def invert_repositories(f, *args, **kwargs):
...
+        repository_ctx.attr.nixopts
+
+    # Large enough integer that Bazel can still parse. We don't have
+    # access to MAX_INT and 0 is not a valid timeout so this is as good
+    # as we can do.
+    timeout = 1073741824
+
+    nix_path = ":".join(
+        [
+            (path_name + "=" + str(repository_ctx.path(target)))
```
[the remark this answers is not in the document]
regnat: That's older than this PR and orthogonal to it, leaving it for later

nixpkgs/nixpkgs.bzl
```diff
@@ -206,6 +206,104 @@ def invert_repositories(f, *args, **kwargs):
...
+            for (target, path_name) in repository_ctx.attr.repositories.items()
+        ],
+    )
```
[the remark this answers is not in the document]
regnat: Same as https://github.com/tweag/rules_nixpkgs/pull/70#discussion_r258563686

nixpkgs/nixpkgs.bzl
```diff
@@ -304,6 +304,79 @@ nixpkgs_packages_instantiate_swapped = repository_rule(
...
+        "--no-build-output",
+        "--add-root", "nix-root", "--indirect",
+        drv_path
+        ] + repository_ctx.attr.nixopts
+    exec_result = _execute_or_fail(
+        repository_ctx,
+        [nix_store_path] + nix_store_args,
+        quiet = True,
+    )
+    output_path = exec_result.stdout.splitlines()[0]
```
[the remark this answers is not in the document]
regnat: Same as https://github.com/tweag/rules_nixpkgs/pull/70#discussion_r258563686 ;)

regnat:
@Profpatsch I think I answered to all your comments apart from https://github.com/tweag/rules_nixpkgs/pull/70#discussion_r258498656, PTAL

nixpkgs/drv_set_builder.nix
```diff
@@ -3,15 +3,22 @@
...
       if builtins.isAttrs x && x ? drvPath && x ? outputName then
       x.drvPath + "!" + x.outputName
-      else null
+      else (evaluationError name).drvPath
       );
     in
     if result.success == true then result.value else null;
 in
-(lib.mapAttrs (name: value: (evaluateElement value)) (packages nixpkgs))
+(lib.mapAttrs (name: value: (evaluateElement name value)) (packages nixpkgs))
```
Profpatsch: nit: can use currying

nixpkgs/nixpkgs.bzl
```diff
@@ -337,6 +337,27 @@ def nixpkgs_packages(
     repository = None,
     nixopts = [],
     ):
+    """
+    Defines a set of targets pointing to nixpkgs packages at once.
```
Profpatsch: pointing to multiple

nixpkgs/nixpkgs.bzl
```diff
@@ -337,6 +337,27 @@ def nixpkgs_packages(
...
+
+    For each `(name, package)` pair in `packages`, this macro will define a
+    `@package` external repository pointing to the nix package `package`
+
```
Profpatsch: “Internally, this calls `nixpkgs_packages_instantiate` once (meaning only one nix evaluation), then `nixpkgs_package_realize` once for each package.”

Profpatsch:
Did we fix the regression that `nix_file` and `attribute_path` should both work at the same time?

nixpkgs/nixpkgs.bzl
```diff
@@ -59,19 +93,196 @@ nixpkgs_local_repository = repository_rule(
...
+        repository_ctx,
+        [ python(repository_ctx)
+        , repository_ctx.path(repository_ctx.attr._json_to_files)
+        , "."
+        , "nix_attrs.nix"
+        ]
+    )
+
+nixpkgs_packages_instantiate_swapped = repository_rule(
+    implementation = _nixpkgs_packages_instantiate_impl,
```
Profpatsch: This should be `_nixpkgs_packages_instantiate_swapped_impl`

nixpkgs/nixpkgs.bzl
```diff
@@ -59,19 +93,196 @@ nixpkgs_local_repository = repository_rule(
...
+    packagesFromFileLabels = \
+      { name: repository_ctx.path(Label(filePath))
+        for (name, filePath) in repository_ctx.attr.packagesFromFile.items()
+      }
+    packagesFromFile = packagesFromFileLabels + generatedPackageFiles
+
+    all_packages = _implode_packages_defs(
+        packagesFromFile,
+        repository_ctx.attr.packagesFromAttr,
+    )
```
Profpatsch: The logic here from line 170 is a bit non-self-explanatory, I feel like it needs some comments.

Might be a good thing to have for the whole subroutine.

nixpkgs/nixpkgs.bzl
```diff
@@ -15,6 +15,28 @@ def python(repository_ctx):
     if python_path:
         return python_path
 
+def _implode_packages_defs(def_files, attribute_paths):
+    """Merge the two dics passed as arguments into a dict of dicts"""
```
Profpatsch: This is a bit more specific than the docstring makes it out to be, it looks like.

nixpkgs/nixpkgs.bzl
```diff
@@ -152,20 +174,29 @@ def _nixpkgs_packages_instantiate_impl(repository_ctx):
...
+    all_packages = _implode_packages_defs(
+        packagesFromFile,
+        repository_ctx.attr.packagesFromAttr,
+    )
+
+    nix_package_defs = \
         [
-            "\"{name}\" = import ''{filePath}'';".format(name = name, filePath = repository_ctx.path(filePath))
-            for (name, filePath) in packagesFromFile.items()
+            # "\"{name}\" = import ''{filePath}'';".format(name = name, filePath = repository_ctx.path(filePath))
```
Profpatsch: nit: stray comment

nixpkgs/nixpkgs.bzl
regnat: I'm not font of exposing the internals in the documentation (not that it's a really hidden internal but just because it's an unnecessary burden for the reader). What about "This is equivalent (but faster) to calling several times `nixpkgs_package`" ?

regnat:
@Profpatsch Had another pass, PTAL

regnat:
@Profpatsch OK to merge this?

regnat:
Mh… you can wait a bit actually, there are ussues with some package names I think

regnat:
Mh… you can wait a bit actually, there are ussues with some package names I think

mboes:
@regnat what's the status on this? Do we still want it?

regnat:
Damn, I though this had been closed already.
Nope, this induces way too much complexity for what it gains us (and it's still not in a 100% working state), so I gave up on it.

I also think the `flakes` thing from nix should give us most of the benefits from this approach once it lands (by allowing the evaluation to be cached), so it's probably even less worth it

PR #83, opened 2019-06-02
Disable implicit nixpkgs config file

guibou:
*This is a breaking change*

By default, nixpkgs will read a global configuration file, by default
`~/.config/nixpkgs/config.nix`. This leads to reproducibility issues
if the configuration is different between users of the repository.

Users  of `nixpkgs_packages` must set the `nix_file_deps` argument
with all the files used by the nix process.

We were previously accepting a global nixpkgs configuration file as
implicit dependency, but this commit now disallow that. User must
explicitly set its own nixpkgs configuration, such as:

```
import nixpkgs_path { config = {}; };
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -36,7 +36,7 @@ nixpkgs_package(
 
 nixpkgs_package(
     name = "expr-test",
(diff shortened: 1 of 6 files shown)


guibou:
@aherrmann Done, thank you.

PR #95, opened 2019-10-31
Define nixpkgs_unix_configure

aherrmann:
Adds `rules_sh` support with `nixpkgs_unix_configure`, a macro to generate a Unix toolchain with commands provided by Nix.

- Update Bazel to version 1.0.0
    `rules_sh` uses the `configure` attribute to `repository_rule` which was only introduced in Bazel 0.29.
- Define `nixpkgs` constraint
    This is intended to move the `nixpkgs` platform constraint defined in `rules_haskell` into `rules_nixpkgs` so that it can be shared. It is used by `nixpkgs_unix_configure` in the `exec_compatible_with` attribute of the generated toolchain.
- Define `nixpkgs_unix_configure`
    This uses `rules_sh`. However, users of `rules_nixpkgs` will only need to depend on `rules_sh` if they use this macro. Otherwise, `rules_nixpkgs` does not depend on `rules_sh`.
- Add tests for `nixpkgs_unix_configure`
    This adds a dependency on `rules_sh` for the test-suite.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -124,9 +124,9 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "rules_sh",
-    sha256 = "ad72710293db59986e1400d70944b3a13a52a11ce411969634669fc545e06c2f",
-    strip_prefix = "rules_sh-5aaea6d44c1ff1ccefc9ec4a22acef301af8f90a",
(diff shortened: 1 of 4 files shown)


Profpatsch:
Can I insist that you move the descriptions into the commits? :)

aherrmann:
> Can I insist that you move the descriptions into the commit messages? :)

:+1:  Done

mboes:
See https://github.com/tweag/rules_sh/issues/1.

aherrmann:
I've updated this PR to use the `rules_sh` release.

aherrmann:
Sounds good, I've renamed everything accordingly.

PR #107, opened 2020-03-07
Added nixpkgs_go_configure

saksmt:
It's a mystery how it wasn't added yet since rules_docker require working go sdk and it doesn't work on NixOS because it's being downloaded as a binary


README.md
```diff
@@ -376,6 +380,94 @@ nixpkgs_cc_configure(repository = "@nixpkgs//:default.nix")
...
 
+### nixpkgs_go_configure
+
+**NOTE: this rule resides in `@io_tweag_rules_nixpkgs//nixpkgs:toolchains/go.bzl` to avoid uncessary dependencies on rules_go for those who don't need go toolchain**
+
+For this rule to work `rules_go` must be available for loading before loading of `@io_tweag_rules_nixpkgs//nixpkgs:toolchains/go.bzl`.
+
+Tells bazel to use go sdk from nixpkgs.
+
+By default rules_go configures go toolchain to be downladed as binaries (which doesn't work on NixOS),
```
mboes:
```suggestion
By default rules_go configures go toolchain to be downloaded as binaries (which doesn't work on NixOS),
```

README.md
```diff
@@ -376,6 +380,94 @@ nixpkgs_cc_configure(repository = "@nixpkgs//:default.nix")
...
+there is a way to tell rules_go to look into environment and find local go binary which is not hermetic.
+This command allows to setup hermetic go sdk from Nixpkgs, which should be considerate as best practice.
+
+Note that nix package must provide full go sdk at the root of pacakage istead of in $out/share/go
```
mboes:
```suggestion
Note that nix package must provide full go sdk at the root of the package instead of in $out/share/go
```

nixpkgs/toolchains/go.bzl
```diff
@@ -0,0 +1,66 @@
...
+        build_file_content = """exports_files(glob(["**/*"]))""",
+        nixopts = nixopts,
+    )
+    
+    if _is_supported_platform:
+        go_wrap_sdk(name = sdk_name, root_file = "@nixpkgs_go_toolchain//:PACKAGE_ROOT")
+
+
+def _is_supported_platform(repository_ctx):
+    return repository_ctx.which("nix-build") != None
```
mboes: missing newline

tests/BUILD.bazel
```diff
@@ -80,3 +82,9 @@ sh_test(
     ],
     toolchains = ["@rules_sh//sh/posix:make_variables"],
 )
+
+# Test nixpkgs_go_configure()
+go_binary(
+    name = "go-test",
+    srcs = ["go-test.go"]
+)
```
mboes: missing newline

regnat:
Really cool :+1: As @mboes said, we currently have a workaround (using `go_version = "host"`), but this is much cleaner.

This breaks the tests however − which unfortunately isn't caught by the CI because it is broken :confused:  − because they still use the very old bazel 1.1 which isn't compatible with the version of `rules_go` used. I tried updating to 2.0 but this breaks the python tests. Maybe we can find something in between so that we can merge this PR quickly and fix the python stuff later

aherrmann:
> I tried updating to 2.0 but this breaks the python tests. Maybe we can find something in between so that we can merge this PR quickly and fix the python stuff later

I took a look at that failure. Fortunately it was easy to work around the issue: https://github.com/tweag/rules_nixpkgs/pull/108

nixpkgs/toolchains/go.bzl
```diff
@@ -0,0 +1,66 @@
...
+        repositories = repositories,
+        nix_file = nix_file,
+        nix_file_deps = nix_file_deps,
+        nix_file_content = nix_file_content,
+        build_file_content = """exports_files(glob(["**/*"]))""",
+        nixopts = nixopts,
+    )
+    
+    if _is_supported_platform:
+        go_wrap_sdk(name = sdk_name, root_file = "@nixpkgs_go_toolchain//:PACKAGE_ROOT")
```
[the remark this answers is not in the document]
regnat: Nvm, `go_local_sdk` can't work since it requires a "static" directory

saksmt:
Pulled latest master, fixed docs (see comment below) and improved nix expression.

Fix in docs:
Unfortunately `go_register_toolchains` uses some sort of reflection/introspection and collects all rules filtering out those that aren't `go_sdk` so we can't just tell users to call `nixpkgs_go_toolchain` (since we can't make it a rule and otherwise we can't find out if nix is present or not) and then `go_register_toolchains` and expect everything to work fine on both nix and non-nix environments, so hacks around rule generation with dynamic loading still needed ([that one](https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/79/files#diff-5ba8ab976de0ec5a2be5606b541d1140) I mean) :slightly_frowning_face: . That means average usefulness of this code drastically drops, but still it's way better then to have go installed on local machine for non-go developer just to be able to build docker images

regnat:
Thanks!

PR #110, opened 2020-03-11
Revert accidental commits

regnat:
Two commits accidentally pushed to master :confused:


aherrmann:
I think we should protect `master` from this happening accidentally. I don't have admin powers on this repo. Could someone who has configure this?

aspiwack:
> I think we should protect `master` from this happening accidentally.

Done this has been. Many minutes ago.

PR #112, opened 2020-03-11
Replace circleci by buildkite

regnat:
Keep the setup synced with the rest of the Tweag repositories

Builds on top of #111 as the CI fails otherwise


mboes:
@regnat fix the author email in the commits and merge?

Issue #114, opened 2020-03-23
More explicit 'Platform is not supported' message?


smelc:
While using rules_nixpkgs for [rules_haskell](https://github.com/tweag/rules_haskell/), I did the following:

```
> mkdir -p test/nested
> cd test
> git clone https://github.com/tweag/rules_haskell
> cd nested
> ./../rules_haskell/start --use-nix # Create rules_haskell standard template
> nix-shell --pure -p bazel
> bazel build //... # Try to build the template
Starting local Bazel server and connecting to it...
ERROR: An error occurred during the fetch of repository 'rules_haskell_sh_posix_nixpkgs':
   Platform is not supported (see 'fail_not_supported')
ERROR: /media/crypt1/repos/test/nested/BUILD.bazel:29:1: //:example depends on @rules_haskell_sh_posix_nixpkgs//:nixpkgs_sh_posix in repository @rules_haskell_sh_posix_nixpkgs which failed to fetch. no such package '@rules_haskell_sh_posix_nixpkgs//': Platform is not supported (see 'fail_not_supported')
ERROR: Analysis of target '//:example' failed; build aborted: no such package '@rules_haskell_sh_posix_nixpkgs//': Platform is not supported (see 'fail_not_supported')
INFO: Elapsed time: 15.039s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (17 packages loaded, 26 targets configured)
    Fetching @rules_haskell_ghc_nixpkgs_haskell_toolchain; Restarting. 12s
```

which is indeed to be expected, because I ran a `--pure` shell with only bazel (so no `nix*` on the `$PATH`).

I was however, puzzled by the `Platform is not supported` message; would it make sense to make explicit that `nix-build` was not found? (as it is what the implementation does: https://github.com/tweag/rules_nixpkgs/blob/master/nixpkgs/nixpkgs.bzl#L59-L60). I can write the patch, but wanted to make sure this is a desirable change beforehand.

aherrmann:
That's a good point.

> would it make sense to make explicit that `nix-build` was not found?

Yes, I think that would be good. It could state that `nix-build` has to be in `PATH` and that `fail_not_supported` can be used if one doesn't want to use Nix.

PR #115, opened 2020-03-25
Make 'platform is not supported' message more detailed

smelc:
This PR fixes issues #114.

### Tests

- Execute the procedure from #114, observe that the detailed error message shows up
- Execute the procedure from #114 but using an impure shell, observe that the build works
- `bazel test //...` at `rules_nixpkgs`'s root works

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -139,7 +139,7 @@ def _nixpkgs_package_impl(repository_ctx):
         expr_args.extend(["-I", nix_path])
 
     if not_supported and fail_not_supported:
-        fail("Platform is not supported (see 'fail_not_supported')")
+        fail("Platform is not supported: nix-build not found in PATH. See attribute fail_not_supported if you don't want to use Nix.")
     elif not_supported:
         return
     else:


nixpkgs/nixpkgs.bzl
```diff
@@ -56,15 +56,26 @@ nixpkgs_local_repository = repository_rule(
...
 
+def _check_supported_platform(repository_ctx):
+    """
+    Returns:
+       None if the platform is supported. A string error message otherwise.
+    """
+    nix_build = repository_ctx.which("nix-build")
+    if nix_build != None:
+        return None
+    return "Platform is not supported: nix-build not found in PATH. See attribute fail_not_supported if you don't want to use Nix."
```
aherrmann: I'm not fond of this approach. This code is surprising and thereby confusing in a couple ways:
* It returns `None` for a positive result
* It returns a non-empty string for a negative result
* It kinda overloads the error message string to function as a boolean at the same time.

I also see no reason for this. The call `repository_ctx.which("nix-build")` is already factored out into `_is_supported_platform`, no reason to factor it out further. The error message is only used in one place, so there's no benefit to factor it out, and even if, the error message is constant, so you could just factor it out into a global binding.

I suggest the following:
* Roll back this change.
* Just edit the error message passed to `fail`.

smelc: I get your point, though in a larger codebase I wouldn't have an error message 50 lines away from the corresponding check, hence my previous solution. But it's also overkill in the current state.

I've edited the commit accordingly.

aherrmann: > I get your point, though in a larger codebase I wouldn't have an error message 50 lines away from the corresponding check, hence my previous solution.

That's a fair point, but in that case I'd go for a solution with less complexity. Maybe something along the lines of

```
def _get_nix_build_or_fail(repository_ctx, fail_not_supported):
    nix_build = repository_ctx.which("nix-build")
    if nix_build == None and fail_not_supported:
        fail(...)
    else:
        return nix_build

def _nixpkgs_package_impl(repository_ctx):
    ...
    nix_build_path = _get_nix_build_or_fail(...)
    if nix_build_path == None:
        return
    ...
```

Issue #85, opened 2019-06-17
Verbose option for nixpkgs_package


guibou:
`nixpkgs_package` is not verbose at all and sometimes users can observe a bazel process stuck on `fetching @some_package` because the underlying nix process takes time. See #82 for a more detailled discussion.

In #82 I proposed to introduce a new attribute to `nixpkgs_package`, `quiet`. This attributes copies the behavior of the similarly named parameter of the `repository_ctx.execute` function: `True` by default, meaning no output (the current behavior), but can be switched `False` by the user, meaning that the output of the underlying subprocess is copied verbatim to bazel outputs.

In this issue, I propose that we implement this attribute.

Note that we cannot just switch back to `quiet = False` by default because it will leaks `nix` output to bazel output in all cases, which may break scripts using, for example, `bazel query`.

This work is blocked by the resolution of #84 because for now `nix-build` is too verbose.

Issue #82, opened 2019-06-02
Verbosity of nixpkgs_package


guibou:
Current `nixpkgs_package` is quiet during package installation. It can be mute as such for a really long time if the nix process is long, giving no user feedback. In case of error, it shows a highly verbose nix output which is totally unusable for the user.

This changes are due to #77, in which  removed the verbosity of `nixpkgs_package`. This was mostly due to #76 which introduced an heuristic for better reproducibility, but this heuristic is using the highly verbose output of `nixpkgs_package` that user do not want to see.

I propose to implement two changes:

- The underlying `nix-build` process will be run twice. Once with nix normal verbosity and the second time with the highest verbosity.
- A `quiet` option will be added to `nixpkgs_package` (and associated rules). It default to `True`, but when switched to `False` it will display the progress of the `nix-build` command, using the normal output.

The heuristic will be computed using the output of the second run. This way, if first run fails, or `quiet` is set to `False`, user will see an informative output. The second run should never fail.

This will introduce a bit of overhead because `nix-build` will be called twice.

# Alternative solutions

- split the build between `nix-instanciate` and `nix-build` and only run the heuristic (hence verbose output) on the `nix-instanciate` call, and setting the `quiet` flag on the `nix-build` call. This will be an improvement in case of error during the `nix-build` phase, however if any error occur during the `nix-instanciate` phase, user will still face a long and verbose output.

Should I implement that?

guibou:
@mboes

guibou:
@Profpatsch @aherrmann

mboes:
I think we're spending too many brain cycles that could be more productively spent elsewhere. I'm not a fan of #76, and not a fan of any of the proposed solutions in this ticket. If we'd just ignore the problem, everything would stay simple. #77 started from a false premise: that Bazel rules are expected to be quiet. I have seen that being said (including by myself), though not substantiated by Bazel documentation, and what applies to build rules need not apply to workspace rules.

Workspace rules are not sandboxed. It would be better if they were (we should file an upstream ticket if there isn't one already), but that's the current state of play. If reimplementing our own mini sandboxing is having knock-on effects, then let's kill it and let the user deal with incorrect workspace rules - they'll be having many others than ours to deal with anyways.

aherrmann:
[...]

PR #118, opened 2020-04-12
Flip back to verbose output in nixpkgs_package

mboes:
Historically, the user was able to see the output of `nix-build` in
real-time. This was useful, because building derivations can take
a long time. Even if the binary cache can be used for all derivations,
downloading can take a long time. Then #77 came along, which made
`nix-build` quiet as a side-effect. Later, #86 subsumed #77 with
a simpler implementation. However, people argued in #82 that being
quiet is the right thing to do, because otherwise workspace rule
output can garble the output of `bazel query` and break scripts. So we
stuck to quiet output and this was called out in the v0.6 changelog.

It turns out that the problem affects other rule authors too (see
bazelbuild/rules_nodejs#583). This led to bazelbuild/bazel#10611,
which fixes the problem for everyone, and shipped in Bazel 3.0. Now,
all workspace rule output goes to `stderr`, so scripts calling `bazel
query` shouldn't be affected.

Given this upstream change, my position is that being verbose is now
the right thing to do. If #85 is implemented, then being verbose
should remain the default, at least for users of Bazel 3.0 onwards.
Because even small packages may have a large set of dependencies that
must be downloaded first. It's nigh impossible for the author of the
workspace file to anticipate the state of the user's cache.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -162,6 +162,7 @@ def _nixpkgs_package_impl(repository_ctx):
             failure_message = "Cannot build Nix attribute '{}'.".format(
                 repository_ctx.attr.attribute_path,
             ),
+            quiet = False,
             timeout = timeout,
         )
         output_path = exec_result.stdout.splitlines()[-1]
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -162,7 +162,7 @@ def _nixpkgs_package_impl(repository_ctx):
             failure_message = "Cannot build Nix attribute '{}'.".format(
                 repository_ctx.attr.attribute_path,
             ),
-            quiet = False,
+            quiet = repository_ctx.attr.quiet,
             timeout = timeout,
         )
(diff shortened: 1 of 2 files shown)


mboes:
I don't understand your first point. Attributes always have defaults.

I don't think we should annoy the majority with a forced upgrade to Bazel 3.0, when it's a small minority that is concerned by `bazel query` output in scripts. All the more so because this PR implements an additional feature to provide them with an easy workaround. And because unlike rules_haskell, rules_nixpkgs does not have a declared Bazel minimum supported version.

nixpkgs/nixpkgs.bzl
```diff
@@ -200,6 +201,7 @@ _nixpkgs_package = repository_rule(
         "build_file": attr.label(),
         "build_file_content": attr.string(),
         "nixopts": attr.string_list(),
+        "quiet": attr.bool(),
```
aherrmann, line 204:
```suggestion
        "quiet": attr.bool(default = False),
```
If I understand @guibou correctly, then his point was to make the default we want explicit.

PR #122, opened 2020-04-19
New configuration setting and platform

### LINKED ISSUE #121: New nix-specific configuration setting

Platforms use the `//nixpkgs/constraints:nixpkgs` constraint value to indicate that they support Nix. However, this constraint value is both poorly named and poorly designed. It's poorly named because we don't care about Nixpkgs support per se, we care about whether Nix is available on the system or not. It's poorly designed because it's a value of the `@bazel_tools//tools/cpp:cc_compiler` constraint setting, which is an abuse of that setting. Platforms can only set one constraint value for any given constraint setting, yet we'd want to set a C/C++ compiler constraint value independently of whether it comes from Nixpkgs or not.

Define a new constraint setting, called `nix`, with one constraint value, called `support_nix`. This is the pattern used in https://github.com/bazelbuild/bazel-toolchains, e.g. for sanitizer support or Docker support.

An alternative would be to ask upstream to define a generic `package_manager` constraint setting, with `nixpkgs` being a constraint value. I ask for precisely this in bazelbuild/platforms#5. Apart from the fact that it has proved to avoid the discussion there from stalling, I think that in the end doing differently may be better anyways. Multiple package managers should be able to coexist. Defining a new `constraint_setting` rather than using a generic one is simpler, works for now, and there isn't much to be gained from defining a unique slot for all package tools like Flatpak, Snappy, Nix etc.

mboes:
Defines a new constraint setting, constraint value, and platform as
per #121. The existing constraint value and platforms are not removed.
Just deprecated.

Closes #121

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,1 @@
-build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/constraints:linux_x86_64_nixpkgs
+build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Changed
 
+- The constraint value for targets to detect whether Nix is available
+  on the current platform is now called
+  `@//nixpkgs/constraints:support_nix`. The associated constraint
+  setting is `@//nixpkgs/constraints:nix`. The old constraint
+  `@//nixpkgs/constraints:nixpkgs` constraint setting is still
+  available. But it is highly recommended to migrate to the new
+  constraint setting, and update platform definitions accordingly.
+  This is a breaking change for users of `nixpkgs_python_configure`.
 - Show Nix output by default, like in releases prior to v0.6.
 
 ## [0.6.0] - 2019-11-14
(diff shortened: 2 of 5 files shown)


Profpatsch:
Sounds fine to me. 

We should have a clear migration description for `rules_nixpkgs` users in the release notes.

mboes:
@aherrmann It does kind of break things, because of the change in the Nixpkgs python stuff. For other rule sets, it won't break anything, unless they also use `nixpkgs_python_configure()`. But they'll need to keep their constraint settings and platforms in sync. You can't change one without changing the other.

An alternative here would be to just not have a deprecation period for the old definitions. It would be a breaking change for more people, but a less subtle one.

aherrmann:
@mboes Can you elaborate how it breaks `nixpkgs_python_configure`?

mboes:
@aherrmann the generated BUILD file in that repository rule states some `exec_with` constraints for the toolchain. It can't state both the old constraint and the new constraint, because then all platforms have to provide both constraint values forever in the future, and in any case that would still be a breaking change. So we move it to the new constraint. Which means that repository rule will only work correctly for users who have updated their platform definition accordingly.

aherrmann:
I see, documenting that well enough in the changelog seems sufficient. I was worried there would be some more fundamental breakage.

PR #123, opened 2020-04-20
Invalid nixpkgs_package on nixpkgs_local_repository change

### LINKED ISSUE #113: nixpkgs_local_repository dependency change doesn't invalidate nixpkgs_package
Given a `WORKSPACE` with something like this:

```bazel
nixpkgs_local_repository(
    name = "nixpkgs",
    nix_file = "//my/nixpkgs/default.nix",
    nix_file_deps = [ "//my/nixpkgs/something.nix"],
)

nixpkgs_package(
    name = "foo",
    attribute_path = "foo",
    repository = "@nixpkgs",
)
```

I'd expect that changes to `my/nixpkgs/something.nix` will force `@foo` to be rebuilt, but it seems that it doesn't and only changes to `my/nixpkgs/default.nix` will do. We had some reproducibility issues because of this, our workaround was to merge everything in one big file and avoid `nix_file_deps`.

From looking quickly through the code, the label `"@nixpkgs"` corresponds to `@nixpkgs//:nixpkgs` and this is symlinked to `my/nixpkgs/default.nix`, so iiuc, from the point of view of `bazel`, `@foo` only depends on that file, and not on `my/nixpkgs/something.nix`. I guess `@nixpkgs//:nixpkgs` could instead be defined to be a `filegroup()` containing the given `nix_file` and everything in `nix_file_deps`?

nlewo:
Additional Nix files of a `nixpkgs_local_repository` are not tracked
by the `nixpkgs_package` rule. In this commit, we generate the list of
these files in the `nixpkgs_local_repository` rule and make the
`nixpkgs_package` rule reading this list to add all of these files in
its own repository to let Bazel track them.

The list of files is propagated from one rule to the other one
throught a file named `nix-file-deps` created in the
`nixpkgs_local_repository` repository.

Closes #113

TODO:
- [ ]  ~write a regression test~ (will be addressed in a dedicated PR)

Thanks @aherrmann for all of his help!

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -26,7 +26,6 @@ nixpkgs_git_repository = repository_rule(
 )
 
 def _nixpkgs_local_repository_impl(repository_ctx):
-    repository_ctx.file("BUILD")
     if not bool(repository_ctx.attr.nix_file) != \
        bool(repository_ctx.attr.nix_file_content):
         fail("Specify one of 'nix_file' or 'nix_file_content' (but not both).")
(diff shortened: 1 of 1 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -36,12 +35,30 @@ def _nixpkgs_local_repository_impl(repository_ctx):
...
     for dep in repository_ctx.attr.nix_file_deps:
-        _cp(repository_ctx, dep)
+        dest = _cp(repository_ctx, dep)
+        dep_files.append(dest)
+
+    # Export all specified Nix files to make them dependencies of a
+    # nixpkgs_package rule.
+    export_files = 'exports_files(["{}", {}])'.format(
+        target_filename.name,
+        ",".join(['"{}"'.format(p) for p in dep_files]))
```
mboes, line 54: A shorter and arguably more readable way of writing this would be,

```python
export_files = "exports_files({})".format(str([target_filename.name] + dep_files))
```

nlewo: Ahah, I agree that's really more readable! Fixed.

PR #126, opened 2020-04-23
Test nixpkgs_package is invalidated on a change in Nix files

nlewo:
This test runs Bazel in the Bazel sandbox to ensure some dependencies
are well tracked at the workspace level. So, we create a workspace,
build it, then change the content of some workspace files and check if
a rebuild of this workspace takes into account these changes.

This is a regression test for the issue #113.


aherrmann:
The test fails for me locally
```
ERROR: /home/aj/.cache/bazel/_bazel_aj/cf7431ac3fb5fc6f4c6a5569d7b55e4a/sandbox/linux-sandbox/1/execroot/io_tweag_rules_nixpkgs/bazel-out/k8-fastbuild/bin/tests/run-test-invalid-nixpkgs-package.runfiles/io_tweag_rules_nixpkgs/BUILD:1:1: //:hello-output depends on @hello//:bin in repository @hello which failed to fetch. no such package '@hello//': Traceback (most recent call last):
        File "/home/aj/.cache/bazel/_bazel_aj/cf7431ac3fb5fc6f4c6a5569d7b55e4a/sandbox/linux-sandbox/1/execroot/io_tweag_rules_nixpkgs/bazel-out/k8-fastbuild/bin/tests/run-test-invalid-nixpkgs-package.runfiles/io_tweag_rules_nixpkgs/nixpkgs/nixpkgs.bzl", line 179
                _execute_or_fail(repository_ctx, <4 more arguments>)
        File "/home/aj/.cache/bazel/_bazel_aj/cf7431ac3fb5fc6f4c6a5569d7b55e4a/sandbox/linux-sandbox/1/execroot/io_tweag_rules_nixpkgs/bazel-out/k8-fastbuild/bin/tests/run-test-invalid-nixpkgs-package.runfiles/io_tweag_rules_nixpkgs/nixpkgs/nixpkgs.bzl", line 622, in _execute_or_fail
                fail(<1 more arguments>)

Cannot build Nix attribute 'hello'.
Command: [/nix/store/jzpx25xjy8lyk21v78k4rhj2ics0g7w9-nix-2.3.3/bin/nix-build, "-E", "import <nixpkgs> { config = {}; overlays = []; }", "-A", "hello", "--out-link", "bazel-support/nix-out-link", "-I", "nixpkgs=/home/aj/.cache/bazel/_bazel_aj/cf7431ac3fb5fc6f4c6a5569d7b55e4a/sandbox/linux-sandbox/1/execroot/io_tweag_rules_nixpkgs/_tmp/e24a3af48b0624ce097203a1f0bab9b3/_bazel_aj/71394a165492de97ddeb25d762a155de/external/nixpkgs/nixpkgs"]
Return code: 1
Error output:
error: could not set permissions on '/nix/var/nix/profiles/per-user' to 755: Read-only file system

ERROR: Analysis of target '//:hello-output' failed; build aborted: no such package '@hello//': Traceback (most recent call last):
        File "/home/aj/.cache/bazel/_bazel_aj/cf7431ac3fb5fc6f4c6a5569d7b55e4a/sandbox/linux-sandbox/1/execroot/io_tweag_rules_nixpkgs/bazel-out/k8-fastbuild/bin/tests/run-test-invalid-nixpkgs-package.runfiles/io_tweag_rules_nixpkgs/nixpkgs/nixpkgs.bzl", line 179
                _execute_or_fail(repository_ctx, <4 more arguments>)
        File "/home/aj/.cache/bazel/_bazel_aj/cf7431ac3fb5fc6f4c6a5569d7b55e4a/sandbox/linux-sandbox/1/execroot/io_tweag_rules_nixpkgs/bazel-out/k8-fastbuild/bin/tests/run-test-invalid-nixpkgs-package.runfiles/io_tweag_rules_nixpkgs/nixpkgs/nixpkgs.bzl", line 622, in _execute_or_fail
                fail(<1 more arguments>)

Cannot build Nix attribute 'hello'.
Command: [/nix/store/jzpx25xjy8lyk21v78k4rhj2ics0g7w9-nix-2.3.3/bin/nix-build, "-E", "import <nixpkgs> { config = {}; overlays = []; }", "-A", "hello", "--out-link", "bazel-support/nix-out-link", "-I", "nixpkgs=/home/aj/.cache/bazel/_bazel_aj/cf7431ac3fb5fc6f4c6a5569d7b55e4a/sandbox/linux-sandbox/1/execroot/io_tweag_rules_nixpkgs/_tmp/e24a3af48b0624ce097203a1f0bab9b3/_bazel_aj/71394a165492de97ddeb25d762a155de/external/nixpkgs/nixpkgs"]
Return code: 1
Error output:
error: could not set permissions on '/nix/var/nix/profiles/per-user' to 755: Read-only file system

INFO: Elapsed time: 9.161s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (5 packages loaded, 5 targets configured)
FAILED: Build did NOT complete successfully (5 packages loaded, 5 targets configured)
```
I'm on Ubuntu 19.10 with Nix single user installation

nlewo:
@aherrmann I reproduced your issue on a Nix single user install. It seems Bazel binds mount the `/nix/store` as read-only in the sandbox. Your Nix processus is then not allowed to write in these directories.

Maybe we could use the Nix `builtins:fetchurl` to fetch an archive containing a directory instead of trying to building it. This way, we don't need to use `nixpkgs` and we could then relocate the Nix store in the Bazel sandbox.

regnat:
> Maybe we could use the Nix builtins:fetchurl to fetch an archive containing a directory instead of trying to building it. This way, we don't need to use nixpkgs and we could then relocate the Nix store in the Bazel sandbox

Or you could directly call `builtins.derivation`, given that it's a trivial example writing the builder manually shouldn't be complicated.

nlewo:
> Or you could directly call `builtins.derivation`, given that it's a trivial example writing the builder manually shouldn't be complicated.

I've been able to create a file with `builtins.derivation`, however `nixpkgs_package` requires a directory as output path. By using `bash` as builder (without `coreutils`), I was not able to create a directory. Do you have something else in mind?

regnat:
Oh that's right, you won't easily have `coreutils` in scope. I though `<corepkgs/utils.nix>` had a reference to it, but it doesn't :thinking: 
Maybe you can use `python` as the builder.

My fear with `fetchurl` is that the test sandbox might prevent you from accessing the network. But maybe that's not the case (which would be a bit creepy though :upside_down_face: )

nlewo:
My plan is to provide an url such as `file://` to `fetchurl`, with a file coming from the Bazel sandbox.

nlewo:
I made a lot of changes and the test is now more hermetic since it no longer uses the Nix deamon (when available).

@aherrmann it should now be passing with a single user Nix installation. However, you have to run:
```
 bazel test //tests:run-test-invalid-nixpkgs-package --sandbox_writable_path /dev/pts
```

regnat:
> @aherrmann it should now be passing with a single user Nix installation. However, you have to run:
> 
> ```
>  bazel test //tests:run-test-invalid-nixpkgs-package --sandbox_writable_path /dev/pts
> ```

I'm quite annoyed by the fact that the test doesn't run out-of-the-box in all the setups. I'd rather disable the sandbox entirely for this test as running it in sandboxed mode doesn't work reliably enough

aherrmann:
I agree with @regnat that it would be preferable to have the tests work without special flags. One can disable the Bazel sandbox on a specific test by setting the [`no-sandox` tag](https://docs.bazel.build/versions/master/be/common-definitions.html#common.tags). Though, then one has to be careful to make the test-case cleanup after itself.

tests/test_invalid_nixpkgs_package.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+set -e
```
aherrmann, line 2:
```suggestion
#!/usr/bin/env bash
set -euo pipefail
```
This script uses bash-isms.

tests/test_invalid_nixpkgs_package.sh
```diff
@@ -0,0 +1,46 @@
...
+export NIX_STATE_DIR=${TEST_TMPDIR}/nix/var/nix
+export NIX_LOG_DIR=${TEST_TMPDIR}/nix/var/log/nix
+export NIX_CONF_DIR=${TEST_TMPDIR}/nix/etc/nix
+
+
+# First, we build a script outputing "hello world 1". This message
+# string comes from the file message.nix which is a dependency of the
+# nixpkgs_package rule.
+echo '"hello world 1"' > message.nix
+bazel build //:hello-output --sandbox_writable_path /dev
```
aherrmann, line 31: Why is the `-sandbox_writable_path /dev` part needed?

tests/BUILD.bazel
```diff
@@ -88,3 +88,18 @@ go_binary(
...
 )
+
+# Note that Nix needs to write to /dev/pts (see
+# https://github.com/NixOS/nix/blob/4a2a45f53d16040f744580747d0963f97ec51164/src/libstore/build.cc#L2255)
+# This test has then to be executed with the flag --sandbox_writable_path /dev/pts
+sh_test(
+    name = "run-test-invalid-nixpkgs-package",
+    timeout = "short",
+    srcs = ["test_invalid_nixpkgs_package.sh"],
+    deps = ["@bazel_tools//tools/bash/runfiles"],
```
aherrmann, line 99:
```suggestion
```
This dependency looks unused. On Unix, paths relative to the runfiles root will work in test-cases. On Windows you'd have to use `rlocation`. However, since `rules_nixpkgs` doesn't work on Windows anyway, it's fine to not use `rlocation`.

nlewo:
A Nix PR removing this `grantpt` call has been merged and I have been able to locally run this test with any special option;)
I will update this PR once this Nix revision have reached the unstable NixOS cache.

tests/test_invalid_nixpkgs_package.sh
nlewo: No, it was actually not required. Sorry.

tests/BUILD.bazel
nlewo: Thanks for explanation.

nlewo:
Tests can now be executed out-of-the-box!

tests/BUILD.bazel
```diff
@@ -88,3 +88,18 @@ go_binary(
     name = "go-test",
     srcs = ["go-test.go"]
 )
+
+# Note that Nix needs to write to /dev/pts (see
+# https://github.com/NixOS/nix/blob/4a2a45f53d16040f744580747d0963f97ec51164/src/libstore/build.cc#L2255)
+# This test has then to be executed with the flag --sandbox_writable_path /dev/pts
```
aherrmann, line 94:
```suggestion
```
As I understand this is no longer true. At least the test passes with a regular `nix-shell --pure --run 'bazel test //...'` for me.

nlewo: Correct. Fixed.

aherrmann: Thank you!

PR #137, opened 2020-08-22
Allow BUILD.pkg to find more kinds of header files

sayrer:
I hit this building an extension for Ruby, which uses .hpp for some files.

### DIFF
--- a/nixpkgs/BUILD.pkg
+++ b/nixpkgs/BUILD.pkg
@@ -12,5 +12,6 @@ filegroup(
 
 filegroup(
     name = "include",
-    srcs = glob(["include/**/*.h"], allow_empty = True),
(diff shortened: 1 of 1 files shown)


nixpkgs/BUILD.pkg
```diff
@@ -12,5 +12,6 @@ filegroup(
 
 filegroup(
     name = "include",
-    srcs = glob(["include/**/*.h"], allow_empty = True),
+    srcs = glob(["include/**/*.h", "include/**/*.hh", "include/**/*.hpp", "include/**/*.hxx"], allow_empty = True),
 )
+
```
aherrmann, line 17:
```suggestion
```

aherrmann (approved):
Thank you!

PR #152, opened 2021-02-10
Include llvm-cov in tools

cocreature:
This seems to be required in Bazel 4.0.0. With this change, CI on the
daml repo is passing on Linux at least (haven’t been able to test on
MacOS for unrelated reasons).


aherrmann-da (approved):
Thank you!

For reference, it was added in https://github.com/bazelbuild/bazel/commit/21b5eb627d78c09d47c4de957e9e6b56a9ae6fad.

aherrmann (approved):

PR #158, opened 2021-03-11
Migrate to GitHub actions & add MacOS pipeline

aherrmann:
depends on #157

- Adds GitHub actions CI workflows for Linux and MacOS (based on the configuration in rules_haskell)
    All tests worked on MacOS right away apart from the one using busybox - the Nix derivation is not-supported on MacOS.
- Adds CI configuration in `.bazelrc`
- Configures BuildBuddy remote cache and build event protocol on CI
- Removes the buildkite configuration
- Updates the Mergify configuration

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,3 +5,18 @@ build --crosstool_top=@nixpkgs_config_cc//:toolchain
 # resolved before `--incompatible_enable_cc_toolchain_resolution` can be
 # recommended for `nixpkgs_cc_configure_hermetic`.
 # build --incompatible_enable_cc_toolchain_resolution
+
+# CI Configuration
+# ----------------
+common:ci --color=no
+build:ci --verbose_failures
+test:ci --test_output=errors
+
+# Use a remote cache during CI
+build:ci --bes_results_url=https://app.buildbuddy.io/invocation/
+build:ci --bes_backend=grpcs://cloud.buildbuddy.io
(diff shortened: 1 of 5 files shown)


mboes (approved):

.bazelrc
```diff
@@ -5,3 +5,22 @@ build --crosstool_top=@nixpkgs_config_cc//:toolchain
...
+
+# CI Configuration
+# ----------------
+common:ci --color=no
+build:ci --verbose_failures
+test:ci --test_output=errors
+
+# Use a remote cache during CI
+build:ci --bes_results_url=https://app.buildbuddy.io/invocation/
+build:ci --bes_backend=grpcs://cloud.buildbuddy.io
```
regnat, line 17: For one second I thought this was remote build. Would have been quite a surprise to see it working

.buildkite/pipeline.yml
```diff
@@ -1,5 +0,0 @@
-steps:
```
regnat, line 1: I think we'll also need to disable buildkite *somewhere* because it currently shows as a failing check in github.

regnat (approved):
Looks like I'm too late and this just got merged, but :+1: anyways

.buildkite/pipeline.yml
aherrmann: Done, I've removed it from the branch protection and archived the pipeline.

PR #159, opened 2021-03-21
Avoid binary file corruption in nixpkgs_local_repository

### LINKED ISSUE #136: nixpkgs_package is corrupting binary files
**Describe the bug**

One of my `nixpkgs_package` rules loads (in the nix file) a binary file. Which ends corrupted when evaluated by nix in the context of `nixpkgs_package` (i.e. after the copy).

**To Reproduce**

Have a binary file directly used in your nix file, something such as:

```
    src = ./abinaryfile.tar.gz;
```

This file must appear in the `nix_file_deps` or your nixpkgs repository. Observes the failure when uncompressing.

**Expected behavior**

It should not corrupt binaries files.

**Environment**
 - OS name + version:
 - Version of the code:

**Additional context**

I tracked the issue to:

https://github.com/tweag/rules_nixpkgs/blob/master/nixpkgs/nixpkgs.bzl#L677

which I extended with the following:

```
+    print(repository_ctx.execute([repository_ctx.which("sha256sum"), repository_ctx.path(src)]).stdout)
+    print(repository_ctx.execute([repository_ctx.which("sha256sum"), repository_ctx.path(dest)]).stdout)
```

And indeed observed that the checksum are not the same.

Aside, I'm not convinced by the `executable=False` flag of the `template` call. It seems to me that a file which is considered executable must keep its bit. I don't know if that's important for nix however.

It seems that `template` does not work well with binary files. Unfortunately there is no such things as a simple `copy` function in repository context.

I tried the following:
    - with an `execute(["cp", ...])` but I ended with a lot of problems due to how `cp` behaves with directory versus files or already existing files
    - with an `symlink`, but issues with existing files.

guibou:
This closes #136

`repository_ctx` is meant to work on "text" files and does text encoding
and decoding, which breaks in the context of a binary file.

There is no `cp` API in repository_ctx, as suggested in
https://github.com/bazelbuild/bazel/issues/11858, but
https://github.com/bazelbuild/bazel/issues/11857 suggested to use this
approach with `repository_ctx.file`.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1210,7 +1210,7 @@ def _cp(repository_ctx, src, dest = None):
     # It means that executable files will miss the bit and won't be executable.
     # Forcing it to True will fix this behavior, but will impact a lot of file
     # (most of the time, files do not have the executable bit) and this may
-    # lead to other errors in the next process are checking for executable bit.
+    # lead to other errors if the next process is checking the executable bit.
     # One other side effect of this change is that you will have a difference
     # in the nix hash computed when nix is run by rules_nixpkgs or directly.
     repository_ctx.file(repository_ctx.path(dest),
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1203,7 +1203,21 @@ def _cp(repository_ctx, src, dest = None):
             for component in [src.workspace_root, src.package, src.name]
             if component
         ])
-    repository_ctx.template(dest, src, executable = False)
+
+    # Copy the file
+    # TODO: auto detect the executable bit of the source file
+    # It is set to False (i.e. not executable) which was the previous behavior.
+    # It means that executable files will miss the bit and won't be executable.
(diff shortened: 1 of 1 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -1203,7 +1203,21 @@ def _cp(repository_ctx, src, dest = None):
...
         ])
-    repository_ctx.template(dest, src, executable = False)
+
+    # Copy the file
+    # TODO: auto detect the executable bit of the source file
+    # It is set to False (i.e. not executable) which was the previous behavior.
+    # It means that executable files will miss the bit and won't be executable.
+    # Forcing it to True will fix this behavior, but will impact a lot of file
+    # (most of the time, files do not have the executable bit) and this may
+    # lead to other errors in the next process are checking for executable bit.
```
regnat, line 1213:
```suggestion
    # lead to other errors if the next process is checking the executable bit.
```

regnat (approved):

aherrmann (approved):
Thank you!

PR #161, opened 2021-03-21
Copy the executable bit

### LINKED ISSUE #160: nixpkgs_local_repository strips executable bit of nix files
**Describe the bug**

`nixpkgs_local_repository` copies the content of all the files listed in its `nix_file_deps` attribute in order to ensure that the build is reproducible.

However, it uses the function `repository_ctx.template(dest, src, executable = False)` or (`repository_ctx.file` in PR #159) which forces the `executable` bit of the copied file to `False`.

A nix process depending on a file with the executable bit set to True will fail.

Setting unconditionally the bit to True or False leads to different behaviors :

- If it is set to `False` (the current behavior), it will change the bit of the (usually few) executable files, making them not executable, which will break the build. I'm having this issue right now.
- If it is set to `True`, the build may fail if a process explicitly checks for the absence of the executable bit, but it is less obvious if this will happen in a real context.

In both cases, changing unconditionally the bit will lead to different nix hash whether if nix is called through rules_nixpkgs or directly. This leads to a duplication of builds.

I don't know how to fix that with the current repository_ctx API which does not provide a way to get file information. Perhaps a combination of `execute` and `read`.

Another option may be to rollback to the former implementation of `nix_file_deps` which was comparing the output of `nix-build -v` with the content of `nix_file_deps`.

guibou:
Closes #160 

Copy the executable bit of files required by `nix_file_deps` in order to ensure that executable binaries can still be executed.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1397,10 +1397,12 @@ def _cp(repository_ctx, src, dest = None):
 
     # Copy the executable bit of the source
     # This is important to ensure that copied binaries are executable.
-    repository_ctx.execute([repository_ctx.which("chmod"),
-                            "--reference",
-                            repository_ctx.path(src),
-                            repository_ctx.path(dest)])
+    repository_ctx.execute([
+        repository_ctx.which("chmod"),
+        "--reference",
+        repository_ctx.path(src),
+        repository_ctx.path(dest),
+    ])
 
     return dest
 
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1388,20 +1388,19 @@ def _cp(repository_ctx, src, dest = None):
         ])
 
     # Copy the file
-    # TODO: auto detect the executable bit of the source file
-    # It is set to False (i.e. not executable) which was the previous behavior.
-    # It means that executable files will miss the bit and won't be executable.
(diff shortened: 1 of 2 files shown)


guibou:
Note that I've based this PR on top of #159 , there are linked.

nixpkgs/nixpkgs.bzl
```diff
@@ -1205,19 +1205,18 @@ def _cp(repository_ctx, src, dest = None):
...
                         executable=False,
                         legacy_utf8=False)
 
+    # Copy the executable bit of the source
+    # This is important to ensure that copied binaries are executable.
+    repository_ctx.execute([repository_ctx.which("chmod"),
+                            "--reference",
+                            repository_ctx.path(src),
+                            repository_ctx.path(dest)])
+
```
aherrmann, line 1219: This change adds a new requirement to `nixpkgs_local_repository`. Before, it only used the `repository_ctx` API and didn't make any further assumptions about the environment. Now, it assumes that `chmod` is in `PATH`. This may not be the case on Windows. A simple way to avoid this issue would be to add a check like the following and simply skip the permission setting on Windows.
```
if get_cpu_value(repository_ctx) == "x64_windows:
```

aherrmann: I've added the proposed check.

aherrmann (approved):

guibou:
Thank you @aherrmann.

PR #163, opened 2021-03-25
Don't impose target constraints on rules_sh

facundominguez:


facundominguez:
Perhaps the note could use some grooming if you have any suggestions.

aherrmann (commented):
There is a stronger distinction between the POSIX toolchain and Bazel's cc toolchain than the note suggests:

The POSIX toolchain provides common shell commands to be executed in build actions as defined by genrules or custom rules. The toolchain should be independent of the target platform in case of cross-compilation, because the output of tools like `find`, or `sort` in a build action is not expected to depend on the target platform.

There is overlap between the tools provided by this toolchain and Bazel's cc toolchain. This overlap is incidental: The list of shell tools defined by the POSIX standard includes some of these tools. However, the POSIX toolchain will not respect the target platform constraint in case of cross-compilation and users should always use Bazel's cc toolchain to compile binaries.

---

The motivation behind setting `target_compatible_with` in the toolchain was for Bazel targets that generate shell scripts that use these tools at runtime. E.g. a test-script that uses `diff` to check whether a program's output is as expected. Clearly, this gets in the way in cross-compilation use-cases.

A good way to accommodate all these different use-cases is to make the platform constraints configurable, both the execute and the target constraints. As a default the choice made in this PR seems very reasonable. I.e. the current platform for the exec constraint and nothing for the target constraint.

The note that users should always use Bazel's cc toolchain to build binaries is probably best placed into the rules_sh documentation. It's not only relevant to rules_nixpkgs implementors but also to users of rules_sh.

aherrmann (approved):
Thank you, I think that explains it well.

PR #165, opened 2021-06-08
change nonexistent nixpkgs_cc_toolchain in readme to nixpkgs_cc_configure

tek:
looks like this was never valid?


aherrmann (approved):
Thank you! Good catch!

aherrmann (approved):

PR #167, opened 2021-08-13
add fail_not_supported to nixpkgs_go_configure

tek:
Passing through the `fail_not_supported` flag to `nixpkgs_package` is trivial, but that alone will cause the `go_wrap_sdk` to fail.
I created another repo rule to wrap that call, in which I use `_is_supported_platform` directly, and which I then call from `nixpkgs_go_configure`.

Problem now is that it seems this repo rule isn't executed at all. Some advice would be appreciated.

Furthermore, I placed the wrapper rule in `nixpkgs.bzl`, because `_is_supported_platform` is private to this file. Is that reasonable?

### DIFF
--- a/nixpkgs/toolchains/go.bzl
+++ b/nixpkgs/toolchains/go.bzl
@@ -7,14 +7,225 @@ dependencies on rules_go for those who don't need go toolchain.
 `@io_tweag_rules_nixpkgs//nixpkgs:toolchains/go.bzl`.**
 """
 
-load(
-    "@io_bazel_rules_go//go:deps.bzl",
-    "go_wrap_sdk",
+load("//nixpkgs:nixpkgs.bzl", "nixpkgs_package")
+load("@bazel_tools//tools/cpp:lib_cc_configure.bzl", "get_cpu_value")
+load("@io_bazel_rules_go//go/private:platforms.bzl", "PLATFORMS")
+
+def _detect_host_platform(ctx):
+    """Copied from `rules_go`, since we have no other way to determine the proper `goarch` value.
+    https://github.com/bazelbuild/rules_go/blob/728a9e1874bc965b05c415d7f6b332a86ac35102/go/private/sdk.bzl#L239
+    """
+    if ctx.os.name == "linux":
+        goos, goarch = "linux", "amd64"
+        res = ctx.execute(["uname", "-p"])
+        if res.return_code == 0:
+            uname = res.stdout.strip()
+            if uname == "s390x":
+                goarch = "s390x"
+            elif uname == "i686":
+                goarch = "386"
+
+        # uname -p is not working on Aarch64 boards
+        # or for ppc64le on some distros
+        res = ctx.execute(["uname", "-m"])
+        if res.return_code == 0:
+            uname = res.stdout.strip()
+            if uname == "aarch64":
+                goarch = "arm64"
+            elif uname == "armv6l":
+                goarch = "arm"
+            elif uname == "armv7l":
+                goarch = "arm"
+            elif uname == "ppc64le":
+                goarch = "ppc64le"
+
+        # Default to amd64 when uname doesn't return a known value.
+
+    elif ctx.os.name == "mac os x":
+        goos, goarch = "darwin", "amd64"
+
+        res = ctx.execute(["uname", "-m"])
+        if res.return_code == 0:
+            uname = res.stdout.strip()
+            if uname == "arm64":
+                goarch = "arm64"
+
+        # Default to amd64 when uname doesn't return a known value.
+
+    elif ctx.os.name.startswith("windows"):
+        goos, goarch = "windows", "amd64"
+    elif ctx.os.name == "freebsd":
+        goos, goarch = "freebsd", "amd64"
+    else:
+        fail("Unsupported operating system: " + ctx.os.name)
+
+    return goos, goarch
+
+go_helpers_build = """
+load("@io_bazel_rules_go//go:def.bzl", "go_sdk")
+
+def go_sdk_for_arch():
+    native.filegroup(
+        name = "libs",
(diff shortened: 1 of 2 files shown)


aherrmann:
Hmm, I see. A problem here is that Bazel's repository rules cannot directly call other repository rules. I.e. calling the repository rule `go_wrap_sdk` (technicall [a function](https://github.com/bazelbuild/rules_go/blob/728a9e1874bc965b05c415d7f6b332a86ac35102/go/private/sdk.bzl#L181-L183) calling a repository rule) from within the implementation of the repository rule `setup_go_sdk` is not legal.

One way to work around this limitation is to have repository rule A generate a `.bzl` file that contains a function that, when called, calls repository rule B as intended. The user then first loads and calls repo rule A and then loads and calls that generated function. Of course, this doesn't really solve the issue here, as the load of B would fail if Nix is not available to build A. I.e. we've just shifted the problem to another place.

A good way to solve this problem may be to avoid using the rules_go repository rules and instead directly call the regular rule that defines the toolchain. Then we can follow a similar pattern as other rules_nixpkgs toolchain setup functions use. E.g. `nixpkgs_python_configure` or `nixpkgs_sh_configure`. The pattern is
1. Call `nixpkgs_package` to build the Go SDK Nix derivation and generate a BUILD file similar to [this template](https://github.com/bazelbuild/rules_go/blob/728a9e1874bc965b05c415d7f6b332a86ac35102/go/private/BUILD.sdk.bazel) (skip the `declare_toolchains` part!)
2. Have a separate repository rule that only depends on the `nixpkgs_package` above via stringly attributes (no label deps!) and generates a BUILD file defining the [`go_toolchain` and `toolchain` rules](https://github.com/bazelbuild/rules_go/blob/728a9e1874bc965b05c415d7f6b332a86ac35102/go/private/go_toolchain.bzl#L123-L143) with a `support_nix` constraint. This indirection makes it so that the `nixpkgs_package` is only depended upon if this Nix provided toolchain is selected in the toolchain resolution.
3. Wrap this all in a Starlark function that calls the `nixpkgs_package` and the second repo rule and then registers the toolchain.

nixpkgs/toolchains/go.bzl
```diff
@@ -7,31 +7,244 @@ dependencies on rules_go for those who don't need go toolchain.
...
 """
 
-load(
-    "@io_bazel_rules_go//go:deps.bzl",
-    "go_wrap_sdk",
+load("//nixpkgs:nixpkgs.bzl", "nixpkgs_package")
+load("@bazel_tools//tools/cpp:lib_cc_configure.bzl", "get_cpu_value")
+load("@io_bazel_rules_go//go/private:platforms.bzl", "PLATFORMS")
+
+def _detect_host_platform(ctx):
```
AleksanderGondek, line 14: I take issue with how this logic is structured.

1. It hides the explicit mappings behind "ifology" statements
2. It takes the "whitelist" approach to supported architectures, thus forcing itself to be always in sync with supported golang architectures. 

I would propose to make the mappings more explicit, extendable and open - for example:

```
BAZEL_OS_TO_GO_OS = {
  "mac os x": "dawin",
}
# This allows us to remember only os names differing from official Bazel names,
# and what is more: pass-thru unexpected names
def get_go_os_name(ctx):
  bazel_os_name = ctx.os.name.lower()
  return BAZEL_OS_TO_GO_OS.get(bazel_os_name, default=bazel_os_name)

BAZEL_ARCH_TO_GO_ARCH = {
  "i686": "386",
  "aarch64": "arm64",
  "armv6l": "arm",
  "armv7l": "arm",
}
# This allows us to remember only arch names differing from official Bazel names,
# and what is more: pass-thru unexpected ones (go may support something we 
# are not yet aware of
def get_go_arch_name(ctx):
  bazel_os_name = ctx.os.name.lower()
  if bazel_os_name in ["linux", mac os x"]:
     
     res = ctx.execute(["uname", "-p"])
     if res.return_code == 0 and ("unknown" not in res.stdout):
       return BAZEL_ARCH_TO_GO_ARCH.get(res.stdout.strip(), default=res.stdout.strip())
       
     res = ctx.execute(["uname", "-m"])
     if res.return_code == 0 and ("unknown" not in res.stdout):
       return BAZEL_ARCH_TO_GO_ARCH.get(res.stdout.strip(), default=res.stdout.strip())
       
     # Pick your poison:
     # fail("Unable to run uname {-p|-m} to detect os arch")
     # return "amd64"
  else:
  # Else assume amd64 because reasons
    return "amd64"
    
    
def _detect_host_platform(ctx):
  get_go_os_name(), get_go_arch_name()
```

nixpkgs/toolchains/go.bzl
```diff
@@ -7,31 +7,244 @@ dependencies on rules_go for those who don't need go toolchain.
...
+nixpkgs_go_helpers = repository_rule(
+    implementation = _nixpkgs_go_helpers_impl
+)
+
+go_toolchain_func = """
+load("@io_bazel_rules_go//go/private:platforms.bzl", "PLATFORMS")
+load("@io_bazel_rules_go//go:def.bzl", "go_toolchain")
+
+def declare_toolchains(host_goos, host_goarch):
+    for p in [p for p in PLATFORMS if not p.cgo]:
```
AleksanderGondek, line 108: This is question coming completely out of ignorance: does this loop mean the cross-compilation capability is always prepared for all known golang PLATFORMS? Is that desired?

AleksanderGondek (commented):
General, non-binding review - overall I just quite dislike the detect_host_platform implementation ;)

aherrmann (approved):
Thank you! This took quite a few hoops to jump through. Good work pushing this through!

I tested it locally and looked at the generated files. Looks good!

Two minor comments, feel free to merge after addressing them.

PR #169, opened 2021-11-23
add exec/target constraints parameters to python toolchain macro

tek:
just copied the code from the CC toolchain, works fine

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -564,19 +564,22 @@ _nixpkgs_cc_toolchain_config = repository_rule(
     },
 )
 
-def _nixpkgs_cc_toolchain_impl(repository_ctx):
+def _ensure_constraints(repository_ctx):
     cpu = get_cpu_value(repository_ctx)
     os = {"darwin": "osx"}.get(cpu, "linux")
-
-    if repository_ctx.attr.target_constraints == None and repository_ctx.attr.exec_constraints == None:
+    if not repository_ctx.attr.target_constraints and not repository_ctx.attr.exec_constraints:
         target_constraints = ["@platforms//cpu:x86_64"]
         target_constraints.append("@platforms//os:{}".format(os))
         exec_constraints = target_constraints
     else:
         target_constraints = list(repository_ctx.attr.target_constraints)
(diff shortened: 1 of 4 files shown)


nixpkgs/nixpkgs.bzl
```diff
@@ -890,6 +890,18 @@ def nixpkgs_cc_configure_deprecated(
...
+    if repository_ctx.attr.target_constraints == None and repository_ctx.attr.exec_constraints == None:
+        target_constraints = ["@platforms//cpu:x86_64"]
+        target_constraints.append("@platforms//os:{}".format(os))
+        exec_constraints = target_constraints
+    else:
+        target_constraints = list(repository_ctx.attr.target_constraints)
+        exec_constraints = list(repository_ctx.attr.exec_constraints)
+
+    exec_constraints.append("@io_tweag_rules_nixpkgs//nixpkgs/constraints:support_nix")
+
```
aherrmann, line 904: This is the same as in `nixpkgs_cc_configure`, correct? Could you factor this out into a function?

aherrmann (approved):
Thank you! Small comment.

CI is failing because the README me is outdated. The test failure should state how to fix it.

nixpkgs/nixpkgs.bzl
tek: with pleasure!

nixpkgs/nixpkgs.bzl
```diff
@@ -890,6 +890,18 @@ def nixpkgs_cc_configure_deprecated(
 
 def _nixpkgs_python_toolchain_impl(repository_ctx):
     cpu = get_cpu_value(repository_ctx)
+    os = {"darwin": "osx"}.get(cpu, "linux")
+
+    if repository_ctx.attr.target_constraints == None and repository_ctx.attr.exec_constraints == None:
```
AleksanderGondek, line 895: You can use `if not variable and not other_variable` to cover uscases of empty strings and null-like values.

tek: empty list is not valid for specification by the consumer?

AleksanderGondek: This is just a soft-ball comment, Starlark is Pythonesque and  I think it is generally a bad smell to use `== None` in Python .

Issue #168, opened 2021-11-15
Add examples of using rules_nixpkgs in a WORKSPACE


roninjin10:
**Is your feature request related to a problem? Please describe.**
When trying to use this package, certain questions came up that we were not able to quickly and easily answer until we tested them out with trial and error.  I believe most of these questions would have been easily and quickly addressed with some example code.

**Describe the solution you'd like**
If we added an example folder that contained a basic example of using rules_nixpkgs, then users of the repo could use the example as a template/starting point for using rules_nixpkgs in their workspace.

PR #170, opened 2021-12-01
Add minimal examples of toolchains in WORKSPACEs

AlexeiDrake:
This addresses issues #168 and #172.


aherrmann (commented):
Nice, thank you!

I think this could use a README with instructions how to build these examples.
It might also be good to add a `.bazelrc` to each example setting the host platform to enable nix support and a `shell.nix` to make these examples more self contained.
We should also test these on CI to make sure that they keep working.

examples/toolchains/cc/README.md
```diff
@@ -0,0 +1,17 @@
+C++ Toolchain Example
+=====================
+
+This is an example C++ project that uses `rules_cc`. It should build and run both with or without Nix installed.
+
+# Usage
+
+To run the example with Nix, issue the following command:
+```
+nix-shell --command 'bazel run :hello'
```
aherrmann, line 10:
```suggestion
nix-shell --command 'bazel run --config nix :hello'
```
Assuming the `build:nix` syntax in `.bazelrc`.

examples/toolchains/cc/README.md
```diff
@@ -0,0 +1,17 @@
...
+```
+
+To run the example without Nix, make sure you have Bazel installed, and issue the following command:
+```
+bazel run :hello
+```
+Note that if this command is run on NixOS, it will still use Nix. :-)
```
aherrmann, line 17: Assuming the `build:nix` syntax in `.bazelrc`, i.e. such that the `--host_platform` is *not* set to the Nix one, this should attempt to use the `local_config_cc` toolchain instead of the Nix provided one. I don't remember if that would work or break on NixOS. That toolchain doesn't download any binaries, it just tries to find tools in `PATH`. If those happen to exist in the given NixOS environment, then this may well still work. But, it's not going to be as hermetic as the Nix provided toolchain.

The same kind of issue holds in the Go case. Though the default Go toolchain will download binaries from the web. If those are statically linked then they may well work on NixOS. But, they would not be Nix provided tools.

examples/toolchains/cc/WORKSPACE
```diff
@@ -0,0 +1,33 @@
...
+load("@io_tweag_rules_nixpkgs//nixpkgs:repositories.bzl", "rules_nixpkgs_dependencies")
+rules_nixpkgs_dependencies()
+
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl",
+     "nixpkgs_git_repository",
+     "nixpkgs_cc_configure"
+)
+nixpkgs_git_repository(
+    name = "nixpkgs",
+    revision = "21.11",
```
aherrmann, line 30: Please also set the `sha256` attribute so that this can be cached in Bazel's repository cache. It's not enabled in this example setup, but it makes for a better example - in the sense of following best practices.

examples/toolchains/cc/shell.nix
```diff
@@ -0,0 +1,2 @@
+{ pkgs ? import <nixpkgs> { } }:
```
aherrmann, line 1: Please pin `nixpkgs`. Otherwise the Bazel version will be floating. Especially important if we want to test this in CI. Otherwise, it's not reproducible.

aherrmann:
Fails on MacOS 11 with 
```
ld: file not found: /usr/lib/system/libcache.dylib for architecture x86_64
clang-5.0: error: linker command failed with exit code 1 (use -v to see invocation)
```
I also see a bunch of errors of the form
```
DEBUG: /private/var/tmp/_bazel_runner/a4a3182e50c9460fa52b8c6c1e6f916f/external/bazel_tools/tools/osx/xcode_configure.bzl:89:14: Invoking xcodebuild failed, developer dir: /Applications/Xcode_12.4.app/Contents/Developer ,return code 1, stderr: src/main/tools/process-wrapper-legacy.cc:80: "execvp(xcrun, ...)": No such file or directory
```

On `rules_haskell` with Nix we [also had linker related issues with MacOS 11](https://github.com/tweag/rules_haskell/runs/3882589783?check_suite_focus=true#step:7:289) (https://github.com/tweag/rules_haskell/issues/1607)
```
Setup.hs: Missing dependency on a foreign library:
* Missing (or bad) C library: z
```
The error looks different at first, but the way Cabal determines whether a library is missing is by [trying to build a C program that links against that library](https://github.com/haskell/cabal/blob/88738ba53070de11052c6ba8df9d1807a92583d4/Cabal/Distribution/Simple/Configure.hs#L1706). Issues with the linker it self could potentially manifest this way as well.

examples/toolchains/cc/hello.cc
```diff
@@ -0,0 +1,6 @@
+#include <iostream>
+
+int main()
+{
+	std::cout << "Hello world!\n";
```
JohnAZoidberg, line 5: Can you please add a dependency to this program, so we know how that would look like?

.github/workflows/workflow.yaml
```diff
@@ -6,11 +6,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-10.15]
+        os: [ubuntu-latest, macos-11]
```
aherrmann, line 9: This update seems to trigger CI to fail with an error that looks similar to https://github.com/tweag/rules_haskell/issues/1607. It's not clear to me at the moment if this is really an issue within rules_nixpkgs or an issue in nixpkgs upstream. Either way, I think this PR shouldn't be blocked on it. It's better to get these examples merged sooner and record the issue to be fixed separately.

Could you switch this one back to the working MacOS version, if the example can't be made to work with MacOS 10.15 then just disable the MacOS test for now, and then open an issue to keep track of this problem.

IIRC the MacOS update was needed because of a more recent nixpkgs needed in the examples because of some compatibility issue. Could you also document this chain of dependencies?

examples/toolchains/cc/README.md
```diff
@@ -0,0 +1,17 @@
+C++ Toolchain Example
+=====================
+
+This is an example C++ project that uses `rules_cc`. It should build and run both with or without Nix installed.
```
aherrmann, line 4: Please clarify that this demonstrates two different modes. Currently, it could be misread that without Nix installed it would somehow install Nix automatically. This is of course not the case, without Nix installed this will fetch toolchains from the environment and be not as hermetic as the Nix setup.

Same for the other `README` files.

examples/toolchains/cc/WORKSPACE
```diff
@@ -0,0 +1,38 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "rules_cc",
+    sha256 = "4dccbfd22c0def164c8f47458bd50e0c7148f3d92002cdb459c2a96a68498241",
+    urls = [
+        "https://github.com/bazelbuild/rules_cc/releases/download/0.0.1/rules_cc-0.0.1.tar.gz",
+    ]
+)
+
+local_repository(
```
aherrmann, line 10: Please add a comment indicating that this should be replaced by an `http_archive` as per the [setup instructions](https://github.com/tweag/rules_nixpkgs/#setup) for a regular use-case outside the rules_nixpkgs repository. That should make it easier for people relatively new to Bazel to understand how to use this.

Same for the other `WORKSPACE` files.

examples/toolchains/cc/WORKSPACE
```diff
@@ -0,0 +1,38 @@
...
+    name = "io_tweag_rules_nixpkgs",
+    path = "../../../"
+)
+
+load("@io_tweag_rules_nixpkgs//nixpkgs:repositories.bzl", "rules_nixpkgs_dependencies")
+rules_nixpkgs_dependencies()
```
aherrmann, line 16: Please run `buildifier` on these files.

examples/toolchains/cc/hello.cc
aherrmann: There are several ways in which a Bazel target could depend on a binary target. E.g. a [`tool` dependency on a `genrule`](https://docs.bazel.build/versions/main/be/general.html#genrule_examples), or as a [`data` dependency on a binary or a test](https://github.com/bazelbuild/codelabs/blob/f07cafe377682d35fd0c82cb816bf0a0fb0c0100/generate_build_files.sh#L112-L115). However, neither of those are specific to rules_nixpkgs, so it seems out of scope for this set of examples.

.github/workflows/workflow.yaml
```diff
@@ -27,3 +27,31 @@ jobs:
...
+          build --remote_header=x-buildbuddy-api-key="$BUILDBUDDY_API_KEY"
+          EOF
+      - name: Build & test
+        run: |
+          for dir in $PWD/examples/toolchains/*; do
+              cd "$dir"
+              echo
+              echo Running $(head -n1 README.md)
+              nix-shell --command 'bazel run --config=nix :hello'
+          done
```
aherrmann, line 57: IIUC the non-Nix use-case is not currently test on CI, correct? Could you add such tests? I guess simply running outside Nix without the nix config should achieve this.
```suggestion
          for dir in $PWD/examples/toolchains/*; do
              cd "$dir"
              echo
              echo Running $(head -n1 README.md)
              nix-shell --command 'bazel run --config=nix :hello'
              bazel run :hello
          done
```

examples/toolchains/cc/WORKSPACE
```diff
@@ -0,0 +1,38 @@
...
+nixpkgs_git_repository(
+    name = "nixpkgs",
+    revision = "21.11",
+    sha256 = "c77bb41cf5dd82f4718fa789d49363f512bb6fa6bc25f8d60902fe2d698ed7cc",
+)
+
+nixpkgs_cc_configure(
+        repository = "@nixpkgs",
+        fail_not_supported = False,
+)
```
aherrmann, line 31: As [mentioned before](https://github.com/tweag/rules_nixpkgs/pull/170#discussion_r761998319) this will name-clash with the autodetecting cc toolchain. You should set `name` explicitly.
```suggestion
nixpkgs_cc_configure(
        name = "nixpkgs_config_cc",
        repository = "@nixpkgs",
        fail_not_supported = False,
)
```

examples/toolchains/python/README.md
```diff
@@ -0,0 +1,13 @@
+Python Toolchain Example
+========================
+
+This is an example Python project with modules that uses `rules_python`. Providing Python modules without Nix is currently not supported.
```
aherrmann, line 4: > Providing Python modules without Nix is currently not supported.

That phrasing seems a bit misleading. This example currently doesn't illustrate the non-Nix setup. However, rules_python is capable of installing Python modules [through pip](https://github.com/bazelbuild/rules_python#installing-pip-dependencies). It's just not as hermetic as the Nix provided ones, because the pip installation runs in a repository rule and is therefore not sandboxed, does not have access to the Python toolchain provided through Bazel's platforms and toolchains, and especially for Python modules with native dependencies could link against arbitrary system installed libraries.

aherrmann (commented):
Thank you! I've done another round of review on this.

Re the MacOS issue, I'd say we should view it as orthogonal to this PR and try to get this PR merged independently of that issue.

examples/toolchains/cc/hello.cc
AlexeiDrake: I've added a Nix-only example in cc_with_deps.

aherrmann: Oh, sorry, misread the request. I thought this was about something depending on the program. This makes sense, thanks @AlexeiDrake for adding it.

aherrmann (approved):
Thank you, that looks great!

I think the remaining CI failure can be fixed by adding a `.bazelignore` file that tells bazel to skip `examples` from the top-level workspace. See [rules_haskell for reference](https://github.com/tweag/rules_haskell/blob/master/.bazelignore)

PR #171, opened 2021-12-10
Add nixpkgs_java_configure

aherrmann:
Add `nixpkgs_java_configure` which imports a Java runtime provided by nixpkgs.
The Java runtime is used by Bazel to execute Bazel built JVM targets, e.g. `java_binary`. It is distinct from the Java toolchain which is used to build JVM targets. Refer to the [Bazel documentation][bazel-java] for details.
This is based on the implementation [in the `daml` repository](https://github.com/digital-asset/daml/pull/11411) (cc @cocreature).

Note, Bazel 4 does not yet support platform based toolchain resolution for the Java toolchain. This feature will only be supported starting from [Bazel 5](https://docs.bazel.build/versions/main/bazel-and-java.html#configuring-the-java-toolchains). Meaning, `nixpkgs_java_configure` is different from e.g. `nixpkgs_cc_configure` in that there is no toolchain registration involved and the user instead has to explicitly pick a particular Java runtime using the `--(host_)javabase` flags.

Adds tests for the nixpkgs provided Java runtime:
- Build and run a `java_test`.
- Test that the selected Java runtime is indeed the Nix provided one.

[bazel-java]: https://docs.bazel.build/versions/4.0.0/bazel-and-java.html#configuring-the-jdk

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,10 +1,16 @@
 build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
+
 build --crosstool_top=@nixpkgs_config_cc//:toolchain
 # Using toolchain resolution can lead to spurious dependencies on
 # `@local_config_cc//:builtin_include_directory_paths`. This needs to be
 # resolved before `--incompatible_enable_cc_toolchain_resolution` can be
 # recommended for `nixpkgs_cc_configure_hermetic`.
 # build --incompatible_enable_cc_toolchain_resolution
+
+build --javabase=@nixpkgs_java_runtime//:runtime
+build --host_javabase=@nixpkgs_java_runtime//:runtime
+build --java_toolchain=@bazel_tools//tools/jdk:toolchain_hostjdk8
+build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_hostjdk8
 
 # CI Configuration
 # ----------------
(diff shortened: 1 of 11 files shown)


.bazelrc
```diff
@@ -1,11 +1,17 @@
...
 # Using toolchain resolution can lead to spurious dependencies on
 # `@local_config_cc//:builtin_include_directory_paths`. This needs to be
 # resolved before `--incompatible_enable_cc_toolchain_resolution` can be
 # recommended for `nixpkgs_cc_configure_hermetic`.
 # build --incompatible_enable_cc_toolchain_resolution
 
+build --javabase=@nixpkgs_java_runtime//:runtime
+build --host_javabase=@nixpkgs_java_runtime//:runtime
+build --java_toolchain=@bazel_tools//tools/jdk:toolchain_hostjdk8
+build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_hostjdk8
```
aherrmann, line 13: We use JDK 8 because the nixpkgs version 18.09 does not provide Java 11 yet and Java 9 and 10 cause problems with Bazel 4.0.0 and are no longer supported releases. See
- bazelbuild/bazel#8750 (comment)
- bazelbuild/bazel#13065

r2r-dev:
Verified it on NixOS unstable. With this change in place we can remove `jdk11` from `shell.nix`.

r2r-dev (approved):

aherrmann:
> Verified it on NixOS unstable. With this change in place we can remove `jdk11` from `shell.nix`.

Good catch, done.

PR #174, opened 2022-01-01
Add stub_shebang to python toolchain py_runtime

JasperDeSutter:
For py_binary rules, bazel creates a stub python script that launches in the system python by default, the stub_shebang allows changing which python is executed.
This should create a more hermetic environment for running python binaries, and fixes bazel version 5 on systems without python3 installed.

https://docs.bazel.build/versions/main/be/python.html#py_runtime_args
https://github.com/bazelbuild/bazel/issues/14159

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1061,7 +1061,16 @@ _nixpkgs_python_toolchain = repository_rule(
     },
 )
 
-_python_nix_file_content = """
+def _python_nix_file_content(attribute_path, bin_path, version):
+    if versions.is_at_least("4.2.0", versions.get()):
+        stub_shebang = """stub_shebang = "#!${{{attribute_path}}}/{bin_path}",""".format(
+            attribute_path = attribute_path,
+            bin_path = bin_path,
+        )
+    else:
+        stub_shebang = ""
+
+    return """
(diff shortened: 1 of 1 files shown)


aherrmann (commented):
Thank you, this is great!

The `stub_shebang` is [a very recent feature](https://github.com/bazelbuild/bazel/commit/763dd0ce6e1644bf895231432f616427a11d385a), though, and not available in Bazel versions <5.0.0.

Could you make the addition of this attribute depend on the Bazel version, to not break compatibility with Bazel 4.x? rules_nixpkgs already depends on Bazel Skylib, so, you should be able to use it's [versions library](https://github.com/bazelbuild/bazel-skylib/blob/6e30a77347071ab22ce346b6d20cf8912919f644/lib/versions.bzl) for this.

JasperDeSutter:
@aherrmann Thanks for the suggestion, I'm only using newer versions so wouldn't have caught that!
I checked the changelog to see the exact version this was added, it seems to date back to [Bazel 4.2.0](https://github.com/bazelbuild/bazel/blob/master/CHANGELOG.md#release-420-2021-08-18). So I've added that version to the check.

aherrmann (approved):
Thank you, looks good!

Good catch on 4.2.0, I missed the cherry-pick.

PR #178, opened 2022-01-18
Add Rust and Foreign CC toolchains

AlexeiDrake:


nixpkgs/toolchains/rust.bzl
```diff
@@ -0,0 +1,159 @@
...
+
+_nixpkgs_rust_toolchain = repository_rule(
+    _nixpkgs_rust_toolchain_impl,
+    attrs = {
+        "toolchain_repo": attr.string(),
+    },
+)
+
+def nixpkgs_rust_configure(
+        sdk_name = "rust_linux_x86_64",
```
aherrmann, line 134:
```suggestion
        name = "nixpkgs_rust_linux_x86_64",
```

Using `name` for consistency with the other rules. Go is the only counter example where `sdk_name` is used for consistency with the Go rules. But, rules_rust doesn't seem to have that convention.

Using `nixpkgs_` prefix to reduce likelihood of collision with a non-Nix repository.

nixpkgs/toolchains/rust.bzl
```diff
@@ -0,0 +1,159 @@
...
+        repository = None,
+        repositories = {},
+        nix_file = None,
+        nix_file_deps = None,
+        nix_file_content = None,
+        nixopts = [],
+        fail_not_supported = True,
+        quiet = False,
```
aherrmann, line 142: Regarding exec and target platform constraints. You could also follow the pattern used e.g. [here](https://github.com/tweag/rules_nixpkgs/blob/05a445575de51872f528c537cc7ff3ab11114f21/nixpkgs/nixpkgs.bzl#L1114-L1115) and then [here](https://github.com/tweag/rules_nixpkgs/blob/05a445575de51872f528c537cc7ff3ab11114f21/nixpkgs/nixpkgs.bzl#L1028) to let the user configure these.

In the simple case [`_ensure_constraint`](https://github.com/tweag/rules_nixpkgs/blob/05a445575de51872f528c537cc7ff3ab11114f21/nixpkgs/nixpkgs.bzl#L569) would pick the current os as the default. That said, aarch64 support is not currently implemented in that approach.

In that case it would make sense to choose a more generic default name like `nixpkgs_rust_config`.

Happy to leave this for a follow-up PR.

nixpkgs/toolchains/foreign_cc.bzl
```diff
@@ -0,0 +1,108 @@
...
+    toolchain_type = "@rules_foreign_cc//toolchains:ninja_toolchain",
+    exec_compatible_with = {exec_constraints},
+    target_compatible_with = {target_constraints},
+)
+"""
+
+def _nixpkgs_foreign_cc_toolchain_impl(repository_ctx):
+    exec_constraints = [ "@io_tweag_rules_nixpkgs//nixpkgs/constraints:support_nix", "@platforms//cpu:x86_64" ]
+    target_constraints = [ "@platforms//cpu:x86_64" ]
+    cpu = get_cpu_value(repository_ctx)
```
aherrmann, line 55: You can use `_ensure_constraints` as shown [here](https://github.com/tweag/rules_nixpkgs/blob/05a445575de51872f528c537cc7ff3ab11114f21/nixpkgs/nixpkgs.bzl#L1028).

aherrmann (commented):
Thank you! Looks good. I've added a couple comments.

examples/toolchains/rust/WORKSPACE
```diff
@@ -0,0 +1,33 @@
...
+)
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+revision = "dc66c1612b7a3e96531eff22136570124c8eec81"
+http_archive(
+    name = "rules_rust",
+    sha256 = "9dadbcd1136f7d4f3f2e7c0790531be0fdcccc535dec42a4c5a6f2df7380e3e3",
+    strip_prefix = "rules_rust-{}".format(revision),
+    urls = [
+        # `main` branch as of 2021-12-01
```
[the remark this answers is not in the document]
ners: Not to my knowledge; this is taken from their setup docs: https://bazelbuild.github.io/rules_rust/#setup
I'll add a note to point to the setup. The last release was in 2018, so I doubt it's a good idea to use it.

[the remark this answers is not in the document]
AlexeiDrake: Not to my knowledge. This is taken directly from the setup: https://bazelbuild.github.io/rules_rust/#setup
I'll add a comment to point to it. Given that the last release was in 2018, I expect this to be the more reliable choice.

examples/toolchains/rust/WORKSPACE
```diff
@@ -0,0 +1,33 @@
...
+
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_git_repository")
+nixpkgs_git_repository(
+    name = "nixpkgs",
+    revision = "21.11",
+    sha256 = "c77bb41cf5dd82f4718fa789d49363f512bb6fa6bc25f8d60902fe2d698ed7cc",
+)
+
+load("@io_tweag_rules_nixpkgs//nixpkgs:toolchains/rust.bzl", "nixpkgs_rust_configure")
+nixpkgs_rust_configure(repository = "@nixpkgs//:default.nix", fail_not_supported = False)
```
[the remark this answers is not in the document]
AlexeiDrake: Good catch, thanks. This is a remnant from before I added bazelrc to each example that allows the user to explicitly specify `config=nix`.
If nix config is not specified, then the toolchain is omitted. But if it is specified, it's certainly better to fail early.

nixpkgs/toolchains/rust.bzl
```diff
@@ -0,0 +1,150 @@
...
+)
+
+filegroup(
+    name = "rust_doc",
+    srcs = ["bin/rustdoc"],
+    visibility = ["//visibility:public"],
+)
+
+load('@rules_rust//rust:toolchain.bzl', 'rust_toolchain')
+rust_toolchain(
```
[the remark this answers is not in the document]
AlexeiDrake: Nice catch. I've added the rest except for rustc_srcs.

nixpkgs/toolchains/rust.bzl
AlexeiDrake: I've added user-configurable parameters to both Rust and foreign-cc toolchains. I'm not sure what you mean by a more generic default name.

nixpkgs/toolchains/rust.bzl
AlexeiDrake: The collision was actually intended: cargo_raze has a hard-coded dependency on `rust_{os}_{arch}//:cargo`, so the easiest way to use it was to register the Nix-provided toolchain with that name. Of course, this can also be user-specified, and as it is cargo_raze specific, I agree that it's probably best to default to a distinct name.

aherrmann (approved):
Good work, thank you!

PR #185, opened 2022-02-16
Support registering a Java toolchain for Bazel 5

cocreature:
My toolchain understanding is very superficial but this seems to work for us on https://github.com/digital-asset/daml and from what I can tell (via --toolchain_resolution_debug) the right toolchain is being selected.

I wasn’t quite sure how to handle backwards compatibility. Doing something magic based on the Bazel version would work I think but being explicit seemed like the nicer option.

### DIFF
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -925,6 +925,41 @@ pkgs.runCommand "bazel-nixpkgs-java-runtime"
   ''
 """
 
+def _nixpkgs_java_toolchain_impl(repository_ctx):
+    cpu = get_cpu_value(repository_ctx)
+    exec_constraints, target_constraints = ensure_constraints(repository_ctx)
+
+    repository_ctx.file(
+        "BUILD.bazel",
+        executable = False,
+        content = """\
+package(default_visibility = ["//visibility:public"])
+
+toolchain(
+    name = "java-toolchain-{cpu}",
+    toolchain = "{runtime}",
+    toolchain_type = "@bazel_tools//tools/jdk:runtime_toolchain_type",
+    exec_compatible_with = {exec_constraints},
+    target_compatible_with = {target_constraints},
+)
+""".format(
+            runtime = repository_ctx.attr.runtime,
+            cpu = cpu,
+            exec_constraints = exec_constraints,
+            target_constraints = target_constraints,
+            version = native.bazel_version,
+        ),
+    )
+
+_nixpkgs_java_toolchain = repository_rule(
+    _nixpkgs_java_toolchain_impl,
+    attrs = {
+        "runtime": attr.label(),
+        "exec_constraints": attr.string_list(),
+        "target_constraints": attr.string_list(),
+    },
+)
+
 def nixpkgs_java_configure(
         name = "nixpkgs_java_runtime",
         attribute_path = None,
@@ -936,10 +971,14 @@ def nixpkgs_java_configure(
         nix_file_deps = None,
         nixopts = [],
         fail_not_supported = True,
-        quiet = False):
+        quiet = False,
+        toolchain = False,
+        exec_constraints = None,
+        target_constraints = None):
     """Define a Java runtime provided by nixpkgs.
 
-    Creates a `nixpkgs_package` for a `java_runtime` instance.
+    Creates a `nixpkgs_package` for a `java_runtime` instance. Optionally,
+    you can also create & register a Java toolchain. This only works with Bazel >= 5.0
     Bazel can use this instance to run JVM binaries and tests, refer to the
     [Bazel documentation](https://docs.bazel.build/versions/4.0.0/bazel-and-java.html#configuring-the-jdk) for details.
 
@@ -976,6 +1015,9 @@ def nixpkgs_java_configure(
       nixopts: See [`nixpkgs_package`](#nixpkgs_package-nixopts).
       fail_not_supported: See [`nixpkgs_package`](#nixpkgs_package-fail_not_supported).
       quiet: See [`nixpkgs_package`](#nixpkgs_package-quiet).
+      toolchain: Create & register a Bazel toolchain based on the Java runtime.
+      exec_constraints: Constraints for the execution platform. Only used when creating a toolchain.
+      target_constraints: Constraints for the target platform. Only used when creating a toolchain.
     """
     if attribute_path == None:
         fail("'attribute_path' is required.", "attribute_path")
@@ -1014,6 +1056,14 @@ def nixpkgs_java_configure(
         fail_not_supported = fail_not_supported,
         quiet = quiet,
     )
+    if toolchain:
+        _nixpkgs_java_toolchain(
+            name = "{}_toolchain".format(name),
+            runtime = "@{}//:runtime".format(name),
+            exec_constraints = exec_constraints,
+            target_constraints = target_constraints,
+        )
+        native.register_toolchains("@{}_toolchain//:all".format(name))
 
 def _nixpkgs_python_toolchain_impl(repository_ctx):
     exec_constraints, target_constraints = ensure_constraints(repository_ctx)
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -922,55 +922,27 @@ pkgs.runCommand "bazel-nixpkgs-java-runtime"
         visibility = ["//visibility:public"],
     )
     EOF
+    cat >$out/java_home.bzl <<EOF
+    java_home = r"${javaHomePath}"
+    EOF
   ''
 """
 
 def _nixpkgs_java_toolchain_impl(repository_ctx):
-    cpu = get_cpu_value(repository_ctx)
-    exec_constraints, target_constraints = ensure_constraints(repository_ctx)
-
     repository_ctx.file(
         "BUILD.bazel",
         executable = False,
         content = """\
-package(default_visibility = ["//visibility:public"])
-config_setting(
-    name = "{name}_name_setting",
-    values = {{"java_runtime_version": "{name}"}},
-    visibility = ["//visibility:private"],
-)
-config_setting(
-    name = "{name}_version_setting",
-    values = {{"java_runtime_version": "{version}"}},
-    visibility = ["//visibility:private"],
-)
-config_setting(
-    name = "{name}_name_version_setting",
-    values = {{"java_runtime_version": "{name}_{version}"}},
-    visibility = ["//visibility:private"],
-)
-alias(
-    name = "{name}_settings_alias",
-    actual = select({{
-        "{name}_name_setting": "{name}_name_setting",
-        "{name}_version_setting": "{name}_version_setting",
-        "//conditions:default": "{name}_name_version_setting",
-    }}),
-    visibility = ["//visibility:private"],
-)
-toolchain(
-    name = "java-toolchain-{cpu}",
-    toolchain = "{runtime}",
-    toolchain_type = "@bazel_tools//tools/jdk:runtime_toolchain_type",
-    exec_compatible_with = {exec_constraints},
-    target_compatible_with = {target_constraints},
-    target_settings = [":{name}_settings_alias"],
+load("@bazel_tools//tools/jdk:local_java_repository.bzl", "local_java_runtime")
+load("@{runtime}//:java_home.bzl", "java_home")
+local_java_runtime(
+   name = "{name}",
+   version = "{version}",
+   runtime_name = "@{runtime}//:runtime",
+   java_home = java_home,
 )
 """.format(
-            runtime = repository_ctx.attr.runtime,
-            cpu = cpu,
-            exec_constraints = exec_constraints,
-            target_constraints = target_constraints,
+            runtime = repository_ctx.attr.runtime_repo,
             version = repository_ctx.attr.runtime_version,
             name = repository_ctx.attr.runtime_name,
         ),
@@ -979,9 +951,7 @@ toolchain(
 _nixpkgs_java_toolchain = repository_rule(
     _nixpkgs_java_toolchain_impl,
     attrs = {
-        "runtime": attr.label(),
-        "exec_constraints": attr.string_list(),
-        "target_constraints": attr.string_list(),
+        "runtime_repo": attr.string(),
         "runtime_version": attr.string(),
         "runtime_name": attr.string(),
     },
@@ -1001,9 +971,7 @@ def nixpkgs_java_configure(
         quiet = False,
         toolchain = False,
         toolchain_name = None,
-        toolchain_version = None,
-        exec_constraints = None,
-        target_constraints = None):
+        toolchain_version = None):
     """Define a Java runtime provided by nixpkgs.
 
     Creates a `nixpkgs_package` for a `java_runtime` instance. Optionally,
@@ -1047,8 +1015,6 @@ def nixpkgs_java_configure(
       toolchain: Create & register a Bazel toolchain based on the Java runtime.
       toolchain_name: The name of the toolchain that can be used in --java_runtime_version.
       toolchain_version: The version of the toolchain that can be used in --java_runtime_version.
-      exec_constraints: Constraints for the execution platform. Only used when creating a toolchain.
-      target_constraints: Constraints for the target platform. Only used when creating a toolchain.
     """
     if attribute_path == None:
         fail("'attribute_path' is required.", "attribute_path")
@@ -1090,9 +1056,7 @@ def nixpkgs_java_configure(
     if toolchain:
         _nixpkgs_java_toolchain(
             name = "{}_toolchain".format(name),
-            runtime = "@{}//:runtime".format(name),
-            exec_constraints = exec_constraints,
-            target_constraints = target_constraints,
+            runtime_repo = name,
             runtime_version = toolchain_version,
             runtime_name = toolchain_name,
         )
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -934,12 +934,11 @@ def _nixpkgs_java_toolchain_impl(repository_ctx):
         executable = False,
         content = """\
 load("@bazel_tools//tools/jdk:local_java_repository.bzl", "local_java_runtime")
-load("@{runtime}//:java_home.bzl", "java_home")
 local_java_runtime(
    name = "{name}",
    version = "{version}",
    runtime_name = "@{runtime}//:runtime",
-   java_home = java_home,
+   java_home = None,
 )
 """.format(
             runtime = repository_ctx.attr.runtime_repo,
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -980,6 +980,8 @@ def nixpkgs_java_configure(
 
     #### Example
 
+    ##### Bazel 4
+
     Add the following to your `WORKSPACE` file to import a JDK from nixpkgs:
     ```bzl
     load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_java_configure")
@@ -997,6 +999,27 @@ def nixpkgs_java_configure(
     # See `bazel query 'kind(java_toolchain, @bazel_tools//tools/jdk:all)'` for available options.
     build --java_toolchain=@bazel_tools//tools/jdk:toolchain_java11
     build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java11
+    ```
+
+    ##### Bazel 5
+
+    Add the following to your `WORKSPACE` file to import a JDK from nixpkgs:
+    ```bzl
+    load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_java_configure")
+    nixpkgs_java_configure(
+        attribute_path = "jdk11.home",
+        repository = "@nixpkgs",
+        toolchain = True,
+        toolchain_name = "nixpkgs_java",
+        toolchain_version = "11",
+    )
+    ```
+
+    Add the following configuration to `.bazelrc` to enable this Java runtime:
+    ```
+    build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
+    build --java_runtime_version=nixpkgs_java
+    build --tool_java_runtime_version=nixpkgs_java
     ```
 
     Args:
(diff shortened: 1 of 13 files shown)


cocreature:
I don’t really know how to write a test for this so I think I need some guidance here.

nixpkgs/nixpkgs.bzl
```diff
@@ -925,6 +925,41 @@ pkgs.runCommand "bazel-nixpkgs-java-runtime"
...
+        content = """\
+package(default_visibility = ["//visibility:public"])
+
+toolchain(
+    name = "java-toolchain-{cpu}",
+    toolchain = "{runtime}",
+    toolchain_type = "@bazel_tools//tools/jdk:runtime_toolchain_type",
+    exec_compatible_with = {exec_constraints},
+    target_compatible_with = {target_constraints},
+)
```
aherrmann, line 944: I was going to suggest to use [`local_java_runtime`](https://github.com/bazelbuild/bazel/blob/41feb616ae18e21fdba3868e4c298b0b83012f10/tools/jdk/local_java_repository.bzl#L41) with `java_home = "", runtime_name = "{runtime}"` following the documentation [here](https://docs.bazel.build/versions/main/bazel-and-java.html#configuring-additional-execution-toolchains). However, I see that `local_java_runtime` doesn't expose `exec|target_compatible_with`, so this doesn't work for us.

I see that `local_java_runtime` generates a couple [`config_setting`s](https://github.com/bazelbuild/bazel/blob/41feb616ae18e21fdba3868e4c298b0b83012f10/tools/jdk/local_java_repository.bzl#L67-L90) to then assign to the `toolchain`'s [`target_setting`](https://docs.bazel.build/versions/main/be/platform.html#toolchain). I haven't followed all the bread crumbs, but I suspect this is needed for the `--java_runtime_version` flag to work.

Perhaps a simple fix is to copy that code to generate those `config_setting`s here as well and raise a ticket on `bazelbuild/bazel` to expose `exec|target_compatible_with` on `local_java_runtime`, so that we can re-use it here in future.

aherrmann (commented):
> I don’t really know how to write a test for this so I think I need some guidance here.

Perhaps adding an example [here](https://github.com/tweag/rules_nixpkgs/tree/de066d554de5bb5fd16270aebb04c99d07d7ed38/examples/toolchains) and testing it with Bazel 5 on CI in a Nix and non-Nix configuration would be a good test. This could even include a test of multiple Java versions...

nixpkgs/nixpkgs.bzl
cocreature: good idea, I’ll try to get to this tomorrow.

cocreature: Tomorrow turned out to be today …

I added the config settings and this does indeed make `--java_runtime_version` work. However, I noticed that `local_java_runtime` also sets up a bunch of default_java_toolchain rules. I’m not entirely sure what this is for but I think one is for compilation while the other is for test execution?

I wonder if duplicating all of this is really a good idea. What do you think of using `local_java_runtime` and accept that it does not set the `target_compatible_with` and `exec_compatible_with`. To differentiate between nix and non-nix platforms you can use `--java_runtime_version` which while being less automatic is also not that terrible and I think at least for the Daml repo would work very well.

cocreature: I pushed a commit for that. Happy to revert if you think that’s a bad idea.

cocreature:
> Perhaps adding an example [here](https://github.com/tweag/rules_nixpkgs/tree/de066d554de5bb5fd16270aebb04c99d07d7ed38/examples/toolchains) and testing it with Bazel 5 on CI in a Nix and non-Nix configuration would be a good test. This could even include a test of multiple Java versions...

I added a test. I’m not sure what you mean by testing a non-nix configuration?

nixpkgs/nixpkgs.bzl
aherrmann: > I added the config settings and this does indeed make `--java_runtime_version` work.

Good to know, thanks for adding and testing!

> However, I noticed that local_java_runtime also sets up a bunch of default_java_toolchain rules. I’m not entirely sure what this is for but I think one is for compilation while the other is for test execution?

Good catch, yes, runtime is for execution (tests and other kinds) and toolchain for compilation.

> To differentiate between nix and non-nix platforms you can use --java_runtime_version which while being less automatic is also not that terrible and I think at least for the Daml repo would work very well.

Are they different? I would have thought, e.g. in the Daml setup, that the versions are the same on Windows and Unix.

aherrmann:
> I added a test. I’m not sure what you mean by testing a non-nix configuration?

Thank you!

E.g. the [C++ example](https://github.com/tweag/rules_nixpkgs/tree/master/examples/toolchains/cc#usage) is setup to also work without Nix (switching via platform setting). It's then tested without Nix [on CI](https://github.com/tweag/rules_nixpkgs/blob/bf6da6adf97c2dae6b477a0f8266c56025fc8cec/.github/workflows/workflow.yaml#L80-L86).

nixpkgs/nixpkgs.bzl
cocreature: The versions are the same but the toolchains are not. Currently the exec and target constraints make sure we select the right one. If we use `local_java_runtime` we need another option to pick the right toolchain depending on the platform. Bazel sets config constraints for `java_runtime_version` for both the name and the version. if you look at the test case. You can then specify either `--java_runtime_version=name`, `--java_runtime_version=version` or `--java_runtime_version=name_version`. So my suggestion is that for the daml repo we would set `build:linux --java_runtime_version=nixpkgs_java` which matches on the name and similar for macos while on Windows, we pick the name passed there which is different.

This isn’t as nice, we really should be able to just set the constraints on the platform. But it seems preferable to me than trying to duplicate both `local_java_runtime` and `default_java_toolchain`.

cocreature:
> E.g. the [C++ example](https://github.com/tweag/rules_nixpkgs/tree/master/examples/toolchains/cc#usage) is setup to also work without Nix (switching via platform setting). It's then tested without Nix [on CI](https://github.com/tweag/rules_nixpkgs/blob/bf6da6adf97c2dae6b477a0f8266c56025fc8cec/.github/workflows/workflow.yaml#L80-L86).

not sure how I managed to miss that, thanks for the pointer! Hopefully this time I’ll actually get to it tomorrow and not tomorrow in a week :)

nixpkgs/nixpkgs.bzl
aherrmann: Ah, I see. Yes, that should work. Indeed, it's not as nice as the platform being sufficient. But, fixing that upstream by exposing those `_compatible_with` attributes seems preferable.

aherrmann:
Thank you!

I tried this out locally. There was a remaining issue that `nixpkgs_package` was required even in the non-Nix use-case. One reason was the `load` of the Nix generated `.bzl` file in the `_toolchain` repository. That can be avoided by leaving `java_home` empty (it's unused when `runtime_name` is set).
Another seems to be that the outcome of `local_java_toolchain` is more strict than seems necessary. This can be worked around by importing a `rules_java` toolchain first.

I've added the needed fixes and verified locally that the generated `java_binary` uses the Nix provided runtime with Nix and the local JDK without Nix. I've also updated the docs a bit.

cocreature:
thanks for the fixes!

aherrmann:
MacOS [is failing](https://github.com/tweag/rules_nixpkgs/runs/5317462794?check_suite_focus=true#step:5:620) with
```
/nix/store/1ym02gx4cmvrxfr1drxfg1dqs9djla86-stdenv-darwin/setup: line 22: shopt: inherit_errexit: invalid shell option name
```
It could be related to https://github.com/NixOS/nixpkgs/issues/155706, but I'm not quite sure why this only happens with the Java example and not the other examples.

AleksanderGondek:
@YorikSar Would you be so kind and look into the macos fail? Sorry to single you out like that, however I feel that you have both recent experience and macos-knowledge to have it debugged in no time :)

cocreature:
> why this only happens with the Java example and not the other examples.

The Java examples use a newer nixpkgs for the nix shell so we get Bazel 5 which probably explains the difference.

nixpkgs/nixpkgs.bzl
```diff
@@ -934,12 +934,11 @@ def _nixpkgs_java_toolchain_impl(repository_ctx):
         executable = False,
         content = """\
 load("@bazel_tools//tools/jdk:local_java_repository.bzl", "local_java_runtime")
-load("@{runtime}//:java_home.bzl", "java_home")
 local_java_runtime(
    name = "{name}",
    version = "{version}",
    runtime_name = "@{runtime}//:runtime",
-   java_home = java_home,
+   java_home = None,
```
cocreature, line 941: I think you can also drop the definition of `java_home.bzl` in that case.

YorikSar:
@cocreature @AleksanderGondek  It looks like the issue I've fixed in https://github.com/tweag/rules_nixpkgs/commit/d3ecfd70a7e2ada4798fa952664d52e54a66e696 - you need to add `nixpkgs.nix` to all examples so that `nix-shell` uses Bash from it instead of the age-old system Bash on macOS. Just copy `nixpkgs.json` and `nixpkgs.nix` from other examples and use it in `shell.nix` in your example to keep it consistent.

aherrmann:
@YorikSar Thank you! That seems to have fixed that issue. I had to use a more recent nixpkgs revision, though, to have Bazel v5 available.
Unfortunately, it's now failing with
```
---8<---8<--- Start of log, file at /private/var/tmp/_bazel_runner/c49486af395adcbbb27603db8559d04f/bazel-workers/multiplex-worker-1-Javac.log ---8<---8<---
Unrecognized VM option 'CompactStrings'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
---8<---8<--- End of log ---8<---8<---
```
AFAIK this issue usually comes up with Java version 8, however, this configuration should be picking version 11...

YorikSar:
If I run `nix-shell --command "bazel run --config=nix :hello"` in `examples/toolchains/java` locally, I get following error:
```
INFO: Invocation ID: 0805d206-be61-4c4a-8cc4-91758a09e32c         
ERROR: /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/local_jdk/BUILD.bazel:2:10: in fail_rule rule @local_jdk//:jdk: 
Traceback (most recent call last):                                  
        File "/private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/rules_java/toolchains/fail_rule.bzl", line 19, column 13, in _fail_rule_impl
                fail("%s %s" % (ctx.attr.header, ctx.attr.message))
Error in fail: Auto-Configuration Error: Cannot find Java binary bin/java in /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/local_jdk/nosystemjdk; either correct your JAVA_HOME, PATH or specify Java from remote repository (e.g. --java_runtime_vers
ion=remotejdk_11                                                  
ERROR: /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/local_jdk/BUILD.bazel:2:10: Analysis of target '@local_jdk//:jdk' failed
ERROR: Analysis of target '//:hello' failed; build aborted:         
INFO: Elapsed time: 0.680s                                          
INFO: 0 processes.                                                  
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
    Fetching @nixpkgs_java_runtime; Restarting.                    
```

It looks like JDK is missing. If I add `pkgs.jdk` to `shell.nix`, it succeeds:

```
 % nix-shell --command "bazel run --config=nix :hello"                 
this path will be fetched (157.28 MiB download, 311.18 MiB unpacked):
  /nix/store/53vc4s0i2dvi3dv6jg0vh930m2rjxcdi-zulu17.30.15-ca-jdk-17.0.1
copying path '/nix/store/53vc4s0i2dvi3dv6jg0vh930m2rjxcdi-zulu17.30.15-ca-jdk-17.0.1' from 'https://cache.nixos.org'...
WARNING: Running Bazel server needs to be killed, because the startup options are different.
Starting local Bazel server and connecting to it...
INFO: Invocation ID: 0cb3bcb9-33b5-4d98-8c73-0b818e128708
this derivation will be built:
  /nix/store/ia8f3v5246y02gkjcm7dk0ps52hkvldd-bazel-nixpkgs-java-runtime.drv
these 3 paths will be fetched (159.47 MiB download, 305.74 MiB unpacked):
  /nix/store/13scxc6gbpmnz98d4bl081w2lvgcglik-stdenv-darwin
  /nix/store/a2v97g7g9dq9vy8v3094lsdgjmwn35yd-adv_cmds-119-locale
  /nix/store/faqx0f36ni8i75xdfsl8g3lsrr4apiv0-zulu11.48.21-ca-jdk-11.0.11
copying path '/nix/store/a2v97g7g9dq9vy8v3094lsdgjmwn35yd-adv_cmds-119-locale' from 'https://cache.nixos.org'...
copying path '/nix/store/faqx0f36ni8i75xdfsl8g3lsrr4apiv0-zulu11.48.21-ca-jdk-11.0.11' from 'https://cache.nixos.org'...
copying path '/nix/store/13scxc6gbpmnz98d4bl081w2lvgcglik-stdenv-darwin' from 'https://cache.nixos.org'...
building '/nix/store/ia8f3v5246y02gkjcm7dk0ps52hkvldd-bazel-nixpkgs-java-runtime.drv'...
/nix/store/y54yn4y7ap3g2yq1ndflgjm5czhiylw7-bazel-nixpkgs-java-runtime
/nix/store/y54yn4y7ap3g2yq1ndflgjm5czhiylw7-bazel-nixpkgs-java-runtime
INFO: Analyzed target //:hello (47 packages loaded, 799 targets configured).
INFO: Found 1 target...
Target //:hello up-to-date:
  bazel-bin/hello.jar
  bazel-bin/hello
INFO: Elapsed time: 104.935s, Critical Path: 9.35s
INFO: 7 processes: 4 internal, 2 processwrapper-sandbox, 1 worker.
INFO: Build completed successfully, 7 total actions
INFO: Build completed successfully, 7 total actions
Hello world
```

Note that even though I added latest `jdk` package with version 17 (`zulu17.30.15-ca-jdk-17.0.1`), rules_nixpkgs downloads version 11 (`zulu11.48.21-ca-jdk-11.0.11`). It looks like something in rules_java initialisation requires JDK being present on machine and uses it over whatever is provided by rules_nixpkgs.

I assume that `macos-11` builder throws different error because it [has](https://github.com/actions/virtual-environments/blob/macOS-11/20220220.1/images/macos/macos-11-Readme.md#java) some Java environment set up, and probably tries to use that one instead (I don't have any Java installed in my environment). Maybe we should add `--pure` to our `nix-shell` calls in CI to make sure we're not affected by that.

YorikSar:
@aherrmann I've tried your new commit. It works, but it spends a bunch of time fetching JDK from the mirror:
```
    Fetching @remotejdk11_macos; fetching 17s
    Fetching https://mirror.bazel.build/openjdk/azul-zulu11.50.19-ca-jdk11.0.12/zulu11.50.19-ca-jdk11.0.12-macosx_x64.tar.gz; 89,865,072B 16s
```
Is this expected?

aherrmann:
@YorikSar Thanks for looking into this! I think the issue is that Bazel treats the compilation and runtime Java toolchains separately. This PR addresses the runtime part, so the JRE used to run Java targets on `bazel run` or `bazel test` or the like.
The compilation part is handled separately. It looks like with the GH actions environment it only finds a JDK 8 unless one adds a JDK 11 to the Nix shell or forces to download a version 11 with the language version flag.

I've looked a bit more closely into `local_java_runtime`. It turns out that it actually defines Java compilation toolchain targets based on the runtime target automatically. So, we could instruct Bazel to use those for compilation as well, which should resolve this issue. Unfortunately, I didn't get this working properly without setting the platform constraints on the generated toolchains.

I've pushed a commit that does this. I've vendored and patched the Java toolchain code from Bazel to achieve this.

I've confirmed locally that this uses a Nix provided JDK for both compilation and execution in Nix-mode and a non-Nix one outside of Nix.

YorikSar:
I've tried this our locally, and while Nix version looks fine, non-nix version fails for me:
```
 % nix shell github:NixOS/nixpkgs/7f9b6e2babf232412682c09e57ed666d8f84ac2d#bazel_5 -c bazel run :hello
INFO: Invocation ID: a89b4270-eaf4-4a12-b5f0-ba80cb6feee5
ERROR: /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/bazel_tools/tools/jdk/BUILD:29:19: While resolving toolchains for target @bazel_tools//tools/jdk:current_java_runtime: no matching toolchains found for types @bazel_tools//tools/jdk:runtime_toolchain_type
ERROR: Analysis of target '//:hello' failed; build aborted: 
INFO: Elapsed time: 0.708s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (3 packages loaded, 5 targets configured)
FAILED: Build did NOT complete successfully (3 packages loaded, 5 targets configured)
```
And adding `jdk11` to the shell didn't help. I guess, something is still using external JDK, even for build.

YorikSar:
It still fails, but with different error:
```
 % nix shell github:NixOS/nixpkgs/7f9b6e2babf232412682c09e57ed666d8f84ac2d#jdk11 github:NixOS/nixpkgs/7f9b6e2babf232412682c09e57ed666d8f84ac2d#bazel_5 -c bazel run --java_runtime_version=remotejdk_11 :hello
INFO: Invocation ID: ef08b488-3641-40b7-b6b7-5e47fc7ece00
ERROR: /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/local_jdk/BUILD.bazel:2:10: in fail_rule rule @local_jdk//:jdk: 
Traceback (most recent call last):
        File "/private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/rules_java/toolchains/fail_rule.bzl", line 19, column 13, in _fail_rule_impl
                fail("%s %s" % (ctx.attr.header, ctx.attr.message))
Error in fail: Auto-Configuration Error: Cannot find Java binary bin/java in /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/local_jdk/nosystemjdk; either correct your JAVA_HOME, PATH or specify Java from remote repository (e.g. --java_runtime_version=remotejdk_11
ERROR: /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/local_jdk/BUILD.bazel:2:10: Analysis of target '@local_jdk//:jdk' failed
ERROR: Analysis of target '//:hello' failed; build aborted: 
INFO: Elapsed time: 0.650s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 3 targets configured)
FAILED: Build did NOT complete successfully (1 packages loaded, 3 targets configured)
```

aherrmann:
@YorikSar Thanks for testing, that's odd. Maybe worth playing with the other Java flags (`--(tool_)java_runtime_version`, `--(tool_)java_language_version` as well.
I see that you're still using a Nix shell even in the "non-Nix" config. Maybe the shell changes something in the environment that confuses the default Java configuration. Have you tried with a bindist version? You can use [bazelisk](https://github.com/bazelbuild/bazelisk) to fetch them by version. E.g. 
```
$ USE_BAZEL_VERSION=5.0.0 bazel info release
release 5.0.0
```

YorikSar:
Yeah, with Bazelisk and `--java_runtime_version=remotejdk_11` it succeeded, even though I didn't install JDK into the env:
```
 % USE_BAZEL_VERSION=5.0.0 bazelisk run --java_runtime_version=remotejdk_11 :hello
2022/02/28 13:42:16 Downloading https://releases.bazel.build/5.0.0/release/bazel-5.0.0-darwin-x86_64...
Extracting Bazel installation...
Starting local Bazel server and connecting to it...
INFO: Invocation ID: c7fb8060-d86b-4d98-8d84-b985c7314dfd
INFO: Build option --java_runtime_version has changed, discarding analysis cache.
INFO: Analyzed target //:hello (10 packages loaded, 707 targets configured).
INFO: Found 1 target...
Target //:hello up-to-date:
  bazel-bin/hello.jar
  bazel-bin/hello
INFO: Elapsed time: 11.966s, Critical Path: 7.13s
INFO: 7 processes: 4 internal, 2 darwin-sandbox, 1 worker.
INFO: Build completed successfully, 7 total actions
INFO: Build completed successfully, 7 total actions
Hello world
```

aherrmann:
@YorikSar Thanks for testing. So, it's probably either a difference in the environment due to Nix shell or a difference in the Nixpkgs provided Bazel version. It looks like nixpkgs [changes `--(tool_)java_runtime_version`](https://github.com/NixOS/nixpkgs/blob/c3f2886c7929e7a16fc26637a793b01a2e7f80f3/pkgs/development/tools/build-managers/bazel/bazel_5/default.nix#L143-L144) compared to their [default values](https://docs.bazel.build/versions/main/bazel-and-java.html#configuring-the-jvm-used-to-execute-and-test-the-code). Perhaps the following would work?
```
 % nix shell \
    github:NixOS/nixpkgs/7f9b6e2babf232412682c09e57ed666d8f84ac2d#jdk11 \
    github:NixOS/nixpkgs/7f9b6e2babf232412682c09e57ed666d8f84ac2d#bazel_5 \
    -c bazel run --tool_java_runtime_version=remotejdk_11 --java_runtime_version=remotejdk_11 :hello
```

YorikSar:
Nope, it still can't find the JDK:
```
Error in fail: Auto-Configuration Error: Cannot find Java binary bin/java in /private/var/tmp/_bazel_tweag/decaa88e167f29f805e634aae0c83deb/external/local_jdk/nosystemjdk; either correct your JAVA_HOME, PATH or specify Java from remote repository (e.g. --java_runtime_version=remotejdk_11
```
It seems like `nix shell` doesn't set `JAVA_HOME` variable at all. `nix shell nixpkgs#jdk11 -c env | grep JAVA` comes up empty. `nix profile install nixpkgs#jdk11` actually does the same. I guess installing Java with Nix is not that straightforward. `home-manager`, for example, adds `JAVA_HOME` explicitly to the shell config: https://github.com/nix-community/home-manager/blob/master/modules/programs/java.nix#L40-L43

aherrmann:
@YorikSar Thanks for testing again. This is pretty odd, it looks like that error [comes from the local toolchain](https://github.com/bazelbuild/rules_java/blob/7a3c520737581f13691ad94a0f798a3518d869d1/toolchains/local_java_repository.bzl#L231-L233). So, despite the `remotejdk_11` flags it seems to select the local one.

That said, the intended configurations for this example, i.e. Nix toolchain with Nix provided Bazel and local/remote toolchain with bindist Bazel. So, I'll merge this one. Could you open a ticket to track the issue you're observing?

aherrmann (approved):

Issue #187, opened 2022-02-17
Rust example fails to compile on macOS because of libiconv


YorikSar:
**Describe the bug**
When I try to build Rust example, I get a link error:

```
...
ERROR: /Users/tweag/workspaces/bazel_on_macos/rules_nixpkgs/examples/toolchains/rust/BUILD:3:12: Compiling Rust bin hello (1 files) failed: (Exit 1): process_wrapper failed: error executing command bazel-out/darwin-opt-exec-FFB32019/bin/external/rules_rust/util/process_wrapper/process_wrapper --subst 'pwd=${pwd}' -- external/nix_rust/bin/rustc hello.rs '--crate-name=hello' '--crate-type=bin' ... (remaining 16 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
error: linking with `external/local_config_cc/cc_wrapper.sh` failed: exit status: 1
  |
[...]

Issue #186, opened 2022-02-17
Python example tries to run as bash script on macOS


YorikSar:
**Describe the bug**
When I run python example, it gets executed as shell script instead of Python with errors like:

```
...
from: can't read /var/mail/__future__
/private/var/tmp/_bazel_tweag/6888388f61d1891f87da729e10be1777/execroot/__main__/bazel-out/darwin-fastbuild/bin/hello: line 13: import: command not found
/private/var/tmp/_bazel_tweag/6888388f61d1891f87da729e10be1777/execroot/__main__/bazel-out/darwin-fastbuild/bin/hello: line 19: del: command not found
...
```

**To Reproduce**
```
cd examples/toolchains/python
nix-shell --command "bazel run --config=nix :hello"
```

**Expected behavior**
"Hello, world!" line should be written to the output.

**Environment**
 - OS name + version: macOS Monterey 12.2.1 (also reproducible on CI with `macos-11` builder
 - Version of the code: Got this issue while trying to make examples work on macOS on [this](https://github.com/tweag/rules_nixpkgs/tree/d3ecfd70a7e2ada4798fa952664d52e54a66e696) commit

**Additional context**
macOS doesn't support calling scripts with scripts in shebang (see [StackOverflow thread](https://stackoverflow.com/questions/67100831/macos-shebang-with-absolute-path-not-working) for example), but python toolchain by default sets shebang as direct path to `python` executable taken from the system. In nixpkgs, that executable is a shell wrapper around `python` itself, so it doesn't work on macOS. It was fixed in NixOS/nixpkgs#93757 for scripts generated by nixpkgs, but this wrapper is generated by Bazel from [this template](https://github.com/bazelbuild/bazel/blob/4.2.2/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt). We already override it for Bazel 4.2.0+ [here](https://github.com/tweag/rules_nixpkgs/blob/de066d554de5bb5fd16270aebb04c99d07d7ed38/nixpkgs/nixpkgs.bzl#L1057), so we should provide full path to Bash before path to Python there.

Issue #176, opened 2022-01-05
Examples don't work on macOS CI runner


AlexeiDrake:
**Describe the bug**
The examples CI job does not work on any version of the macOS runner.

**To Reproduce**
Add `macos-11` or `macos-10.15` to the list of runners for the examples job.

**Expected behavior**
CI jobs should pass.

**Additional context**
1. Our examples pin nixpkgs, ideally to a release to avoid cache misses. At the time of writing, the most obvious options are 21.05 and 21.11.
2. The Go toolchain requires at least Bazel 4.2.x., which is present in 21.11, but not in 21.05 (4.0.0).
3. We must therefore pin against at least 21.11.
4. nixpkgs at 21.11 seems to have broken the Darwin stdenv for macos-10.15 with this commit: https://github.com/NixOS/nixpkgs/commit/a634fbe065c1800ebb1e1e1dd7a691115280493d
5. We must therefore use at least macos-11.
6. macos-11 CI support is broken and tracked with this bug: [tweag/rules_nixpkgs#175](https://github.com/tweag/rules_nixpkgs/issues/175)

aherrmann:
Thanks for writing up the context!

> [...] ideally to a release to avoid cache misses [...]
> 3\. We must therefore pin against at least 21.11.

This is not strictly true. Nixpkgs is continuously built on hydra and pushed to the nixpkgs community cache in between releases as well. To find a cached nixpkgs revision that provides that bazel version you can
* browse to the [nixpkgs hydra job](https://hydra.nixos.org/jobset/nixpkgs/trunk)
* [search](https://hydra.nixos.org/jobset/nixpkgs/trunk#tabs-jobs) for [bazel_4 jobs](https://hydra.nixos.org/job/nixpkgs/trunk/bazel_4.x86_64-darwin/all)
* find one that provides Bazel version 4.2.x but predates https://github.com/NixOS/nixpkgs/commit/a634fbe065c1800ebb1e1e1dd7a691115280493d.
* and lookup the nixpkgs revision in its [inputs](https://hydra.nixos.org/build/153268139#tabs-buildinputs)

[...]

PR #188, opened 2022-02-17
macOS CI fixes

### LINKED ISSUE #175: CI jobs fail on macOS 11
**Describe the bug**
CI jobs fail on macOS 11.

**To Reproduce**
Add the `macos-11` runner to the CI workflow.

**Expected behavior**
CI jobs should pass.

**Additional context**
This update seems to trigger CI to fail with an error that looks similar to tweag/rules_haskell#1607. It's not clear at the moment if this is really an issue within rules_nixpkgs or an issue in nixpkgs upstream.

YorikSar:
* bump nixpkgs to 21.11 everywhere
* switch to `macos-11` builder tests
* enable `macos-11` builder for examples, excluding Python and Rust ones (see #186 and #187)

This should close #175 and #176.


YorikSar:
For some reason GitHub expects build that is removed in this PR (the `macos-10.15` one).

.github/workflows/workflow.yaml
```diff
@@ -61,6 +77,18 @@ jobs:
           for dir in cc go rust; do
               cd "$dir"
               echo
+              if [ "${{ runner.os }}" == "macOS" ]; then
+                case "$dir" in
+                  # Python - https://github.com/tweag/rules_nixpkgs/issues/186
+                  # Rust - https://github.com/tweag/rules_nixpkgs/issues/187
+                  python | \
```
AleksanderGondek, line 84: python will never be the case in this loop, we can safely omit it

.github/workflows/workflow.yaml
```diff
@@ -53,6 +57,18 @@ jobs:
           for dir in *; do
```
AleksanderGondek, line 57: I am worried about the increasingly nested, loop, if-else driven structure of this script.
Perhaps we can think about improving it's readability.
For example:
```
...
bazel_run_with_nix() {
  echo Running $(head -n1 README.md) with Nix
  nix-shell --command 'bazel run --config=nix :hello'
  cd ..
}

bazel_run_without_nix() {
  echo Running $(head -n1 README.md) without Nix
  bazel run :hello
  cd ..
}

bazel_skip() {
  echo Skipping $(head -n1 README.md)
  cd ..
}

runner_os="${{ runner.os }}" 
for dir in *; do
  cd "${dir}"
  echo
  case "${dir},${runner_os}" in
    python,macOS)
      bazel_skip
      continue
      ;;
    rust,macOS)
      bazel_skip
      continue
      ;;
    cc,*)
      bazel_run_with_nix
      bazel_run_without_nix
      ;;
    go,*)
      bazel_run_with_nix
      bazel_run_without_nix
      ;;
    rust,*)
      bazel_run_with_nix
      bazel_run_without_nix
      ;;
    *,*)
      bazel_run_with_nix
      ;;
  esac
done
```

AleksanderGondek (changes requested):
Great work on pinning the issues with macOS!
However I would really like to avoid making the test execution script a highly nested, if-else driven construct – I have given a suggestion in review, but I do not mind any solution ;)

.github/workflows/workflow.yaml
```diff
@@ -32,7 +34,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest] # TODO: this job does not yet work on any version of macOS
+        os: 
+          - ubuntu-latest # TODO: this job does not yet work on any version of macOS
```
AleksanderGondek, line 38: Perhaps we can remove the comment, seems like it no longer hold strue.

YorikSar: Good point, will do.

.github/workflows/workflow.yaml
YorikSar: I just wanted to make sure we don't have different logic in different places here.

.github/workflows/workflow.yaml
YorikSar: I'm more afraid of building a complicated bash-based test matrix runner than just nesting and if. Maybe I could make it a bit simpler? I'll try.

YorikSar: Please check out the new version - it seems even more straightforward.

.github/workflows/workflow.yaml
YorikSar: Fixed in the new version.

AleksanderGondek (approved):
LGTM

aherrmann:
> For some reason GitHub expects build that is removed in this PR (the `macos-10.15` one).

This was a branch protection setting. I've changed it to point to macos-11

PR #189, opened 2022-02-21
Fix python example build on macOS

### LINKED ISSUE #186: Python example tries to run as bash script on macOS

YorikSar:
Add `bash` if necessary to the `#!` line of the Python scripts. Leverage makeScriptWriter from nixpkgs that includes fix NixOS/nixpkgs#93757 for macOS compatibility.

Fixes #186

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -64,8 +64,6 @@ jobs:
           without_nix=( cc go rust )
 
           if [ "${{ runner.os }}" == "macOS" ]; then
-            # Python - https://github.com/tweag/rules_nixpkgs/issues/186
-            skip with_nix python
             # Rust - https://github.com/tweag/rules_nixpkgs/issues/187
             skip with_nix rust
             skip without_nix rust
--- a/nixpkgs/nixpkgs.bzl
+++ b/nixpkgs/nixpkgs.bzl
@@ -1053,16 +1053,15 @@ _nixpkgs_python_toolchain = repository_rule(
 )
 
 def _python_nix_file_content(attribute_path, bin_path, version):
-    if versions.is_at_least("4.2.0", versions.get()):
-        stub_shebang = """stub_shebang = "#!${{{attribute_path}}}/{bin_path}",""".format(
-            attribute_path = attribute_path,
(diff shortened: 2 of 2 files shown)


AleksanderGondek (approved):
LGTM

PR #193, opened 2022-03-07
extract `bzlmod` modules for core rules and toolchains

### LINKED ISSUE #182: Split rules_nixpkgs into separate components
Requirement of #181 

The new [bzlmod][bzlmod] dependency management requires rule sets to declare all their direct dependencies upfront. 
`rules_nixpkgs` is in the situation that it depends on many other rule sets. In principle, it could depend on every Bazel extension that provides a compiler toolchain like `rules_cc`, `rules_go`, `rules_python`, etc. 

Before bzlmod this was not a problem, because dependencies were lazy - in the sense that e.g. `nixpkgs_go_configure` depends on `rules_go`, but a user only needs to worry about that dependency when they actually load `@io_tweag_rules_nixpkgs//nixpkgs:toolchains/go.bzl`.

With `bzlmod` all these dependencies need to be declared upfront in the `MODULE.bazel` file with version constraints. Meaning, `rules_nixpkgs`, as is, would introduce a large set of transitive dependencies for any of its users.

This is problematic as it could unnecessarily inflate the set of transitive dependencies of a project using `rules_nixpkgs` and expose users to version conflicts in transitive dependencies that they don't even want to use. It's also problematic as it would block modularization of `rules_nixpkgs` until all of its dependencies have been successfully modularized.

The issue can be avoided by splitting `rules_nixpkgs` into separate Bazel modules, one for every language integration.

This could take the following form:
- `rules_nixpkgs_core`:
  Defines `nixpkgs_git_repository`, `nixpkgs_local_repository`, and `nixpkgs_package`;
  and exposes other common functionality.
- `rules_nixpkgs_cc`:
  Defines `nixpkgs_cc_configure`, etc.
- `rules_nixpkgs_java`:
  Defines `nixpkgs_java_configure`.
- etc.
- We can keep a top-level `rules_nixpkgs` that re-exposes all these components for compatibility with the legacy non-bzlmod workflow.

Note, the intent is not to create separate Git repositories for all these components - this should not be necessary. It should still be possible to maintain all these components within the same Git repository.

[bzlmod]: https://docs.bazel.build/versions/5.0.0/bzlmod.html

fricklerhandwerk:
closes #182

This is a fairly large change in terms of LOC, and also has some noise due to a generated `README.md` being checked into version control. It's best to follow it commit by commit, as they are fairly self-contained.

The main line of thought is to literally separate rules by concern, as described in #182, and keep the tests working (which accounts for most of the non-obvious changes).


nixpkgs/repositories.bzl
```diff
@@ -27,3 +33,17 @@ def rules_nixpkgs_dependencies():
         url = "https://github.com/bazelbuild/rules_java/releases/download/4.0.0/rules_java-4.0.0.tar.gz",
         sha256 = "34b41ec683e67253043ab1a3d1e8b7c61e4e8edefbcad485381328c934d072fe",
     )
+    if not local:
+        maybe(
+            http_archive,
+            "rules_nixpkgs_core",
+            # XXX: no way to use `sha256` here, but if this surrounding repo comes
+            # from that URL, Bazel should hit the cache for the sub-workspace
```
aherrmann: I'm not sure it will, AFAIK Bazel uses the sha256 as the cache key.

While looking into this I found the following: [`native.existing_rule`](https://docs.bazel.build/versions/main/skylark/lib/native.html#existing_rule) exposes the attribute of previously defined (repository) rules as well. I think we could construct something like the following (this is only a very rough sketch)
```
def rules_nixpkgs_dependencies(*, rules_nixpkgs_url = "", rules_nixpkgs_sha256 = ""):
    # *_url|sha256 attributes for fallback/user override
    if rules_nixpkgs_url:
        # Use user override ...
    rules_nixpkgs = native.existing_rule("io_tweag_rules_nixpkgs")
    if rules_nixpkgs:
        kind = rules_nixpkgs.get("kind")
        if kind == "local_repository":
            path = rules_nixpkgs.get("path")
            maybe(
                local_repository,
                "rules_nixpkgs_core",
                path = path + "/core",
            )
        elif kind == "http_archive":
            url = rules_nixpkgs.get("url")
            urls = rules_nixpkgs.get("urls")
            sha256 = rules_nixpkgs.get("sha256")
            strip_prefix = rules_nixpkgs.get("sha256")
            # ... other relevant attributes, e.g. patches?
            maybe(
                http_archive,
                "rules_nixpkgs_core",
                url = url,
                urls = urls,
                sha256 = sha256,
                strip_prefix = strip_prefix + "/core",
                ...
            )
        else:
            # Fall back to user provided override or fail if missing
```
(Again, only a very rough sketch and missing a couple edge-cases and error handling)

toolchains/rust/MODULE.bazel
```diff
@@ -4,3 +4,8 @@ module(
 )
 
 bazel_dep(name = "rules_nixpkgs_core", version = "0.8.1")
+# TODO: there is no BCR entry for `rules_go` yet, and you will have to add a
```
aherrmann, line 7: Typo `rules_go` --> `rules_rust`.

docs/BUILD.bazel
```diff
@@ -17,6 +17,18 @@ stardoc(
     deps = ["//nixpkgs"],
 )
 
+stardoc(
+    name = "core",
```
aherrmann, line 21: Shouldn't there also be stardocs and generated READMEs for the other components?
Side note: Happy to leave the docs aspect for a follow-up PR.

nixpkgs/BUILD.bazel
```diff
@@ -59,6 +62,7 @@ bzl_library(
     deps = [
         ":bazel_tools",
         ":nixpkgs",
+        "//toolchains/go:go.bzl",
```
aherrmann, line 65: I'm a bit puzzled. This is only the backwards compatibility file, right? I.e. the one that reexports. Shouldn't there be a dependency on `@rules_nixpkgs_go` somewhere?

aherrmann (commented):
Great work! Thank you!
A few comments, though this already looks very good.

To be sure that the backwards compatibility aspect works we should also test this on some users of rules_nixpkgs. rules_haskell and daml come to mind.

nixpkgs/toolchains/rust.bzl
```diff
@@ -1,5 +1,5 @@
 load("@rules_nixpkgs_core//:nixpkgs.bzl", "nixpkgs_package")
-load("//nixpkgs:private/constraints.bzl", "ensure_constraints")
+load("//core:util.bzl", "ensure_constraints")
```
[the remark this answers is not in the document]
fricklerhandwerk: This was fixed in subsequent commits.

nixpkgs/BUILD.bazel
fricklerhandwerk: No, this points to the new module (the old file would be `//nixpkgs/toolchains:go.bzl`). The target is only used for docs generation, and indeed appears a bit misplaced. For some reason all the other language specifics had been in `nixpkgs` and only `go` was separate. I would clean this up step by step, for example when separating out tests in #194. This PR is already huge.

aherrmann: Thanks for clarifying. Yes, happy to defer this to a follow-up PR.

nixpkgs/repositories.bzl
fricklerhandwerk: Thanks for the hint. Implemented it in a reasonably general fashion.

docs/BUILD.bazel
fricklerhandwerk: Prepared that step and opened #197.

docs/BUILD.bazel
```diff
@@ -69,6 +79,41 @@ genrule(
...
+        name = "readme_{}".format(rule),
+        srcs = [
+            "README_{}.md.tpl".format(rule),
+            "toolchains/{}.md".format(rule),
+        ],
+        outs = ["toolchains/{}/README.md".format(rule)],
+        # sorry for this mess. Bazel `str.format()` does not allow nested `{}`
+        # pairs and `awk` is sensitive to line breaks.
+        # TODO: consider alternative: leave reference for the end and just
+        # append the generated string with `cat`.
```
aherrmann, line 93: I think one can escape `{` as `{{` and then it can be nested.

examples/toolchains/rust/WORKSPACE
```diff
@@ -1,4 +1,3 @@
-# Replace with http_archive: https://github.com/tweag/rules_nixpkgs/#setup
```
aherrmann, line 1: I think this can stay. At least it did in the other examples.

aherrmann (approved):
Thank you! Good working teasing those apart!

A few last minor comments, and https://github.com/avdv/daml/commit/04dff55df40ec51b69dac9974436e70ba242d225#r68737946. Feel free to merge once they're resolved.

This is one of those PRs, where git's `colorMoved = default` and `colorMovedWS = allow-indentation-change` is really useful.

aherrmann:
Uhm, that was a bit faster than intended, sorry about that! It looks like we lost the `merge-queue` label requirement in the Mergify config. I've opened https://github.com/tweag/rules_nixpkgs/pull/199 to fix that.
@fricklerhandwerk In that case, could you address the last minor comments in a follow-up?

avdv:
> To be sure that the backwards compatibility aspect works we should also test this on some users of rules_nixpkgs. rules_haskell and daml come to mind.

FTR, here are the changes that were necessary in the daml project to use rules_nixpkgs after extracting the modules:

https://github.com/digital-asset/daml/compare/main...avdv:rules_nixpkgs-pr-193

(due to not using `rules_nixpkgs_dependencies()` the logic from `nixpkgs/repositories.bzl` had to be duplicated)

docs/BUILD.bazel
fricklerhandwerk: 🤔 Does that make *anything* substantially better? I'll try...

fricklerhandwerk: It does not work, not even in Python 3.8 (`ValueError: unexpected '{' in field name`). This is a pathological case. Just look at the amount of additional braces! It is probably better to have a simpler setup in the first place, as noted in the comment.

avdv:
> @avdv Which patch do you mean?

In daml they are applying a patch to rules_nixpkgs to avoid "occasional segmentation faults of nix":
https://github.com/digital-asset/daml/blob/e0965709fe62b5c86f980ef25472a5eeb59f3adb/deps.bzl#L51-L59

this one:
https://github.com/digital-asset/daml/blob/main/bazel_tools/nixpkgs-disable-http2.patch

I did not care too much about it, as I did not see a segfault and the patch could be adapted if needed.

aherrmann:
@avdv Thanks for testing and sharing the diff!

> due to not using rules_nixpkgs_dependencies() the logic from nixpkgs/repositories.bzl had to be duplicated

Raised a question [here](https://github.com/digital-asset/daml/commit/cb4f8749f64854055a0a576c6797e11deda7c937#r68817729):
> Was it necessary to duplicate this here, or could this have been done by calling `rules_nixpkgs_dependencies` in `WORKSPACE` as well?

PR #198, opened 2022-03-15
use `coreutils` for sandbox tests

fricklerhandwerk:
now all tests also run on Darwin.


aherrmann (approved):
Nice! Good catch, thank you!

PR #200, opened 2022-03-16
Additional changes after bzlmod separation

avdv:
This PR rounds out the changes in #193:

* add `BUILD.pkg` file back
* handle `strip_prefix` of `git_repository` and `http_archive` when adding the sub-modules


aherrmann (approved):
Good catch, thank you!

PR #202, opened 2022-03-16
Separate generating documentation per module

### LINKED ISSUE #197: Generate reference documentation for each toolchain module
As part of the effort described #181, generated reference documentation should be available per module. Right now we only generate it for the cumulative `rules_nixpkgs`.

Fix the `TODO`s outlined in #193: simplify and move generator rules into `rules_nixpkgs_core`, use them in each toolchain module, and also in the legacy `WORKSPACE`.

fricklerhandwerk:
closes #197

this is a huge diff, unfortunately. most of the noise comes from unavoidable redundancy in `WORKSPACE` files per module for backwards compatibility, and the rendered documentation checked into version control. tried to keep a narrative in commit history. it probably makes sense to just look at the result, possibly excluding the last commit, which only updates the `README`s. to inspect motivation per line one has to resort to `git blame`, sorry for the inconvenience.

key takeaways:
- documentation is rendered per module
- rendering code is shrinked and simplified significantly, and in one place
- top-level `README.md` renders from the same data
- full backwards compatibility


docs/README.md
```diff
@@ -0,0 +1,1423 @@
...
+  name = "local-nixpkgs",
+  path = "/path/to/nixpkgs",
+)
+
+nixpkgs_package(
+  name = "somepackage",
+  repositories = {
+    "nixpkgs": "@local-nixpkgs//:default.nix",
+  },
+  â¦
```
aherrmann, line 106:
```suggestion
```
Not sure where this comes from. It'll have to be removed at the source.

docs/README.md
```diff
@@ -0,0 +1,1423 @@
...
+
+* [nixpkgs_git_repository](#nixpkgs_git_repository)
+* [nixpkgs_local_repository](#nixpkgs_local_repository)
+* [nixpkgs_package](#nixpkgs_package)
+* [nixpkgs_cc_configure](#nixpkgs_cc_configure)
+* [nixpkgs_cc_configure_deprecated](#nixpkgs_cc_configure_deprecated)
+* [nixpkgs_java_configure](#nixpkgs_java_configure)
+* [nixpkgs_python_configure](#nixpkgs_python_configure)
+* [nixpkgs_go_configure](#nixpkgs_go_configure)
+* [nixpkgs_rust_configure](#nixpkgs_rust_configure)
```
aherrmann, line 36: Looks like Go and Rust are missing. At least their links are not active.

core/.bazelrc
```diff
@@ -0,0 +1,35 @@
+build --host_platform=@rules_nixpkgs_core//platforms:host
+
+# NOTE: all of the following is copy-pasted from `/.bazelrc` and necessary to
+# make this repository work compatibly in isolation.
```
aherrmann, line 4: Perhaps the common parts could be placed in a `.bazelrc.common` and then `import`ed by both `core/.bazelrc` and `/.bazelrc`?

core/WORKSPACE
```diff
@@ -1,2 +1,67 @@
-# only temporary for compatibility with `WORKSPACE` setup
-# TODO: remove when migration to `bzlmod` is complete
\ No newline at end of file
+# NOTE: only temporary for compatibility with `WORKSPACE` setup!
+# TODO: remove when migration to `bzlmod` is completelocal_repository
```
aherrmann, line 2:
```suggestion
# TODO: remove when migration to `bzlmod` is complete
```

core/WORKSPACE
```diff
@@ -1,2 +1,67 @@
...
+
```
aherrmann, line 3: Best add a comment to explain that the following are only dev-dependencies needed for stardoc.

core/BUILD.bazel
```diff
@@ -1,33 +1,23 @@
-load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("//docs:stardoc.bzl", "generate")
```
aherrmann, line 1: Unfortunately, this is problematic.

Separating the Stardoc parts into `//docs` means that users of `rules_nixpkgs_core` who do not depend on anything in `@rules_nixpkgs_core//docs` do not require Stardoc. In other words, moving the Stardoc dependency from `//docs` into `//` lifts Stardoc from being just a dev-dependency into being a regular dependency.

But, Stardoc is not needed for users of `rules_nixpkgs_core`, so, it should just be a dev-dependency.

core/BUILD.bazel
```diff
@@ -1,33 +1,23 @@
...
 )
 
-bzl_library(
-    name = "bazel_tools",
-    srcs = [
-        "@bazel_tools//tools:bzl_srcs",
-    ],
-)
-
-bzl_library(
```
aherrmann, line 22: The `bzl_library` targets are needed for other rule sets depending on `rules_nixpkgs_*`.

E.g. `rules_haskell` generates its API docs using Stardoc, too. But, Startdoc requires all transitively loaded `.bzl` files to be declared `bzl_library` dependencies. So, to generate `rules_haskell` API docs, we'll need the `rules_nixpkgs` `bzl_library` targets.

toolchains/java/WORKSPACE
```diff
@@ -1,2 +1,62 @@
...
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@io_tweag_rules_nixpkgs//nixpkgs:repositories.bzl", "rules_nixpkgs_dependencies")
+
+rules_nixpkgs_dependencies()
+
+load("@rules_nixpkgs_core//:nixpkgs.bzl", "nixpkgs_git_repository", "nixpkgs_local_repository")
+load("//:java.bzl", "nixpkgs_java_configure")
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_cc_configure")
+
+http_archive(
```
aherrmann, line 23: Perhaps the docs related parts could be factored out into `@rules_nixpkgs_core//docs:repositories.bzl` or something along those lines?

core/docs/stardoc.bzl
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+def copy_files(name, data):
+    """copy list of files to workspace root"""
+    native.sh_binary(
+        name = name,
+        srcs = ["@rules_nixpkgs_core//docs:copy-files.sh"],
+        args = ["$(location {}) {}".format(a, b) for a, b in data],
+        data = [a for a, b in data],
+        env = {"PATH": "$(POSIX_CP)"},
```
aherrmann, line 107: Does that really do the intended thing? I would expect this to set `PATH=/nix/store/.../bin/cp`, whereas I would expect us to want `PATH=/nix/store/.../bin`. Am I missing something?

toolchains/cc/README.md
```diff
@@ -0,0 +1,231 @@
+<!-- Generated with Stardoc: http://skydoc.bazel.build -->
+
+<!-- Edit this docstring and run `bazel run //docs:update-README.md` to change this repository's `README.md`. -->
```
aherrmann, line 3:
```suggestion
<!-- Edit the docstring in `toolchainc/cc/cc.bzl` and run `bazel run //docs:update-README.md` to change this repository's `README.md`. -->
```
Given that this comment will be copied into the generated `README.md` it may be better if it points to the actual location of the docstring to be edited.

aherrmann (commented):
Thank you, good work teasing all of this apart!

A couple comments.

core/BUILD.bazel
aherrmann: I see that the `bzl_library` is generated as part of the docs `generate` macro. As mentioned in https://github.com/tweag/rules_nixpkgs/pull/202#discussion_r831945136, we'll need the `bzl_library` targets in a BUILD file that does not depend on Startdoc, so that down-stream (e.g. rules_haskell) can define its own `bzl_library` targets, without depending on dev-dependencies of `rules_nixpkgs_*`.

core/docs/stardoc.bzl
fricklerhandwerk: It did work as intended from a vanilla environment, but now that you say it, it surprises me, too. Will re-check.

core/BUILD.bazel
fricklerhandwerk: I was under the impression that those `bzl_library` targets are _only_ used for documentation. Files like `@rules_nixpkgs_core//:nixpkgs.bzl` are exposed by filegroup `:srcs` if I understand correctly. One can still run tests and examples when I delete all calls to `generate()`. Also I can't find where `rules_haskell` may depend on the `nixpkgs` `bzl_library`.

core/BUILD.bazel
fricklerhandwerk: I see what you mean, but I'm not sure this is really a problem here. There is no Bazel code path that would even go from dependants into `stardoc.bzl` unless explicitly called.

core/.bazelrc
fricklerhandwerk: Even easier, just add another symlink.

docs/README.md
fricklerhandwerk: This is very weird. I will have to pull apart the docs-specific repository rules first, to see what the difference to the other toolchains even is. They build individually, and I wouldn't know why in this collection anything would change.

core/BUILD.bazel
aherrmann: When another project references a target in this BUILD file, e.g. the `bzl_library` target, then Bazel will load this BUILD file. It will also try to load the `load` commands, which will lead it to depend on Stardoc, because `stardoc.bzl` `load`s it. `load` cannot be guarded, so, this makes it a strict dependency.

fricklerhandwerk: What do you suggest? Add a `doc` package for each module and do `generate()` in there?

core/docs/stardoc.bzl
fricklerhandwerk: Yes, `$PATH` is indeed augmented with `/nix/store/.../bin/cp` and therefore ignored. The rest of `$PATH` still has a minimal `nix` environment including `coreutils`, but it seems not to come from `nixpkgs_git_repository`: changing the `nixpkgs` release there does not influence the store paths. That is weird, because I test this in
```
bash
unset NIX_PATH
unset NIX_PROFILES
unset NIX_PROFILE_DIR
export PATH=/nix/store/<...>-nix-2.3.16/bin:/usr/bin:/usr/sbin:/bin:/sbin
/nix/store/<...>-bazel-4.2.1/bin/bazel run //:update-README.md
```

I don't really see how one would actually take the proper `bin` path to `cp` without tinkering within the script. I would like to keep the script absolutely minimal and not concern itself with these environmental questions.

aherrmann: It's defined in the Nix expression for Bazel itself here: https://github.com/NixOS/nixpkgs/blob/19d6e349b910b19d54b58a5b3af14ee0e1efedea/pkgs/development/tools/build-managers/bazel/bazel_4/default.nix#L78-L126

docs/README.md
fricklerhandwerk: Ah, we did not have them in `//nixpkgs:nixpkgs.bzl` to avoid strict dependencies on Go and Rust tooling, and so there are no aliases. Alright, this goes hand with the other changes.

fricklerhandwerk:
@aherrmann Fixed all the comments and issues.

docs/README.md
aherrmann: The Go one seems to be still missing, it's not contained in the toplevel `README`. [The diff](https://github.com/tweag/rules_nixpkgs/pull/202/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L1411) shows it as removed.

I noticed that the symbol is missing in the list in `docs/BUILD.bazel`, but the following patch didn't change things:
```
diff --git a/docs/BUILD.bazel b/docs/BUILD.bazel
index 24e46ed..6a14150 100644
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -11,6 +11,7 @@ generate_documentation(
         "nixpkgs_cc_configure_deprecated",
         "nixpkgs_java_configure",
         "nixpkgs_python_configure",
+        "nixpkgs_go_configure",
         "nixpkgs_rust_configure",
         "nixpkgs_sh_posix_configure",
     ],
```

aherrmann (commented):
Good work, this looks great!

Only one comment: https://github.com/tweag/rules_nixpkgs/pull/202#discussion_r835136727

docs/README.md
fricklerhandwerk: Yes, this is intended, see https://github.com/tweag/rules_nixpkgs/commit/d55b95f9672b4ebaf6540abf6c69a873e2ece4f6. If we generated Go in the top-level `README` we either would have to alias `nixpkgs_go_configure` in `nixpkgs/nixpkgs.bzl`, which would force Go dependencies on unsuspecting users; or complicate the build machinery disproportionately. I decided to just link to the submodule.

aherrmann: I see, makes sense to go this route.

aherrmann (approved):

PR #211, opened 2022-03-29
Add ::group:: for each example to GitHub workflow logs

YorikSar:
It should make it easier to find the exact example in logs. See run for this commit for an example: https://github.com/tweag/rules_nixpkgs/runs/5736918187


tek (approved):

avdv:
TIL :+1:

Issue #180, opened 2022-01-26
Support remote execution with rules_nixpkgs


aherrmann:
**Is your feature request related to a problem? Please describe.**
Bazel supports [remote execution](https://docs.bazel.build/versions/main/remote-execution.html) through the [remote execution protocol](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto). This protocol manages inputs and outputs required by and generated by Bazel build actions, i.e. actions defined by [regular Bazel rules](https://docs.bazel.build/versions/main/skylark/rules.html).

However, `rules_nixpkgs` defines [repository rules](https://docs.bazel.build/versions/main/skylark/repository_rules.html) that invoke `nix-build` during Bazel's loading phase. The Nix package manager will then realize Nix store paths (typically under `/nix/store/...`) and generate symlinks into Bazel's execution root. These Nix store paths are outside of Bazel's control and the remote execution protocol does not ensure that these store paths are also realized on the remote execution nodes.

Remote execution actions that depend on Nix store paths will fail if the required Nix store paths are not realized on the remote execution nodes.

**Describe the solution you'd like**
We need some solution to ensure that Nix store paths that are required for Bazel build actions exist on the remote execution nodes that these actions may be run on.

Some possible approaches:
* Run a script before each Bazel invocation (e.g. by defining a custom [`tools/bazel`](https://github.com/bazelbuild/bazel/issues/5788)) that builds all required Nix store paths on the remote execution nodes.
* Remote execution nodes could share a Nix store via network fs.
* `nixpkgs_package` could set [`remotable = True`](https://docs.bazel.build/versions/5.0.0/skylark/lib/globals.html#repository_rule(implementation,%20attrs,%20local,%20environ,%20configure,%20remotable,%20doc)) to execute the `nix-build` command on the remote execution nodes. ([feature announcement](https://groups.google.com/g/bazel-discuss/c/xm0wgsCzTG0/m/d20QwsbrBQAJ), [commit](https://github.com/bazelbuild/bazel/commit/06a26c37126e3eefabd28c24a4e35059b5b136bc), [flag](https://docs.bazel.build/versions/main/command-line-reference.html#flag--experimental_repo_remote_exec))
[...]

PR #214, opened 2022-04-03
Rule to generate docker images with nix store paths

jcpetruzza:
**Context**: I'm currently using bazel with remote execution using this branch; it's not an ideal solution, but it's a first step and there could be other people interested in trying it out.

The goal was to have a docker image that contains all the nix store paths that occur in the workspace, so that the remote executors also inherit them. For this, I'm introducing a repository rule `nixpkgs_docker_image`, that I'm then using in my workspace as follows:

```bzl
nixpkgs_docker_image(
    name = "nix_deps_image",
    srcs = [
        "@cc_toolchain_nixpkgs_info//nixpkg",
        "@nixpkgs_python_toolchain_python3//nixpkg",
        "@nixpkgs_sh_posix_config//nixpkg",
        "@rules_haskell_ghc_nixpkgs//nixpkg",
        "@nixpkgs_valgrind//nixpkg",
    ],
    bazel = "@nixpkgs_bazel//nixpkg",
    repositories = {"nixpkgs": "@nixpkgs"},
)
```

In `srcs`, one should include every usage of the `nixpkgs_package` rule, either explicit (as `@nixpkgs_valgrind` in this example) or implicit as those coming from the toolchain rules. One then gets two targets:

  - `@nix_deps_image//:image`: a `genrule` that uses `dockerTools.buildLayeredImage` to build a tarball
  - `@nix_deps_image//:stream`: a `sh_binary` that uses `dockerTools.streamLayeredImage` to output the tarball on stdout.

One then needs to ensure that the docker image is available before starting a remote build, either by manually pushing the image, or running a CI step.

All this relies on the `nixpkgs_package` rule creating a `nixpkg` subpackage in the package repository, containing the exact nix-expression used by `nix-build` to create the package. Ideally, this would all be more transparent to the user, who would just mention the toolchain and the package in `srcs` and the rule would figure out the rest, but for a proof-of-concept this is enough.

The funny bit is in the `bazel` argument, what's this about? There is an unexpected issue coming from using bazel provisioned by nixpkgs (which I guess is the most common case for `rules_nixpkgs` users): internal bazel files like `@bazel_tools//tools/cpp:linux_cc_wrapper.sh.tpl` get the shebang mangled so that `#!/usr/bin/env bash`, becomes, say `#!/nix/store/5nr4gffsb5l72xpz6d5jbd0skih3r9i2-bash/bin/bash` and the problem is how to ensure that those paths ends up in the docker image as well. The workaround here is for the user to optionally provide a `nixpkgs_package` that should correspond to the exact same bazel being used on dev, CI, etc., so that whatever was used when mangling shebangs is available as a dependency.

As far as I can see, this issue goes beyond `rules_nixpkgs`, in that a nixpkgs-provisioned bazel is hard to use with remote execution and may require some upstream fix. I think all the possible approaches listed in #180 would be affected by this.

### DIFF
--- a/core/default.nix.tpl
+++ b/core/default.nix.tpl
@@ -1,0 +1,12 @@
+{%{args_with_defaults}}:
+let
+  x = %{def};
+  y = if builtins.isFunction x then
+        let
+          formalArgs = builtins.functionArgs x;
+          actualArgs = builtins.intersectAttrs formalArgs {inherit %{args};};
+        in
+           x actualArgs
+       else x;
+in
+  y%{maybe_sel}
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -36,6 +36,7 @@ load(
     "expand_location",
     "find_children",
     "is_supported_platform",
+    "label_to_path",
 )
 
 def _nixpkgs_git_repository_impl(repository_ctx):
@@ -124,9 +125,25 @@ Create an external repository representing the content of Nixpkgs, based on a Ni
 """,
 )
 
+def _nixopts_args(nixopts):
+    result = {}
+    arg_opt, arg_name = None, None
+    for opt in nixopts:
+        if opt == "--arg" or opt == "--argstr":
+            arg_opt, arg_name = opt, None
+        elif arg_opt:
+            if arg_name == None:
+                arg_name = opt
+            else:
+                arg_val = opt if arg_opt == "--arg" else "''%s''" % opt
+                result[arg_name] = arg_val
+                arg_opt, arg_name = None, None
+    return result
+
 def _nixpkgs_package_impl(repository_ctx):
     repository = repository_ctx.attr.repository
     repositories = repository_ctx.attr.repositories
+    attribute_path = repository_ctx.attr.attribute_path
 
     # Is nix supported on this platform?
     not_supported = not is_supported_platform(repository_ctx)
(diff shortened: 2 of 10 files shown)


dmadisetti:
@aherrmann Regarding #291 , does this look like a reasonable approach? Wondering why this was never reviewed / commented on

core/default.nix.tpl
```diff
@@ -0,0 +1,12 @@
...
+  x = %{def};
+  y = if builtins.isFunction x then
+        let
+          formalArgs = builtins.functionArgs x;
+          actualArgs = builtins.intersectAttrs formalArgs {inherit %{args};};
+        in
+           x actualArgs
+       else x;
+in
+  y%{maybe_sel}
```
aherrmann, line 12: Maybe more descriptive names help make this easier to understand?

```suggestion
  value_or_function = %{def};
  value = if builtins.isFunction value_or_function then
        let
          formalArgs = builtins.functionArgs value_or_function;
          actualArgs = builtins.intersectAttrs formalArgs {inherit %{args};};
        in
           value_or_function actualArgs
       else value_or_function;
in
  value%{maybe_sel}
```

core/nixpkgs.bzl
```diff
@@ -124,9 +125,25 @@ Create an external repository representing the content of Nixpkgs, based on a Ni
...
+    result = {}
+    arg_opt, arg_name = None, None
+    for opt in nixopts:
+        if opt == "--arg" or opt == "--argstr":
+            arg_opt, arg_name = opt, None
+        elif arg_opt:
+            if arg_name == None:
+                arg_name = opt
+            else:
+                arg_val = opt if arg_opt == "--arg" else "''%s''" % opt
```
aherrmann, line 154: Could this trigger quoting issues if a users passes in a quoted `--argstr` option?

core/nixpkgs.bzl
```diff
@@ -162,22 +179,56 @@ def _nixpkgs_package_impl(repository_ctx):
     if repository_ctx.attr.nix_file and repository_ctx.attr.nix_file_content:
         fail("Specify one of 'nix_file' or 'nix_file_content', but not both.")
     elif repository_ctx.attr.nix_file:
-        nix_file = cp(repository_ctx, repository_ctx.attr.nix_file)
-        expr_args = [repository_ctx.path(nix_file)]
+        src = repository_ctx.attr.nix_file
+        cp(repository_ctx, src, label_to_path(src, prefix="nixpkg/src"))
```
aherrmann, line 199: Could you explain what this change does and why it's needed? IIUC this introduces an extra prefix and nests the copied files deeper. Wouldn't that break cross workspace location expansion references?

aherrmann (commented):
@dmadisetti Thank you for the ping! @jcpetruzza sorry for the super late reply, this fell of the plate.

> Regarding https://github.com/tweag/rules_nixpkgs/issues/291 , does this look like a reasonable approach?

It's targetting a different issue, namely that of providing all needed build dependencies for remote execution. It could also be used for deployables, if the deployable you're looking for are Docker images, and if you don't mind potentially an over-approximation of your runtime dependencies.

@jcpetruzza regarding the remote execution aspect. How does this approach fare related to the incrementality concerns raised in #180. I.e. if one Nix package is changed and the image needs to be rebuilt. Doesn't this invalidate all build actions and effectively force a full rebuild of the project?

cc @benradf @aloiscochard

jcpetruzza:
Hi @aherrmann!  

> regarding the remote execution aspect. How does this approach fare related to the incrementality concerns raised in https://github.com/tweag/rules_nixpkgs/issues/180. I.e. if one Nix package is changed and the image needs to be rebuilt. Doesn't this invalidate all build actions and effectively force a full rebuild of the project?

Good question; I guess it depends on whether bazel considers the container-image of a (host) platform definition part of the build input or not, but I don't really know what it does here. My setup follows [this guide](https://www.buildbuddy.io/docs/rbe-platforms), so essentially I have something like this:
```
_NIX_DEPS_IMAGE_TAG = "f3279413...."

platform(
    name = "nix_deps_image_platform",
    constraint_values = [
        "@bazel_tools//platforms:x86_64",
        "@bazel_tools//platforms:linux",
        "@bazel_tools//tools/cpp:clang",
    ],
    exec_properties = {
        "OSFamily": "Linux",
        "container-image": "docker://some.registry/nix-deps-image:%s" % _NIX_DEPS_IMAGE_TAG,
    },
    parents = [
        "@io_tweag_rules_nixpkgs//nixpkgs/platforms:host",
    ],
)
```

And to use RBE I then do something like:

```
bazel build --remote_executor=grpcs://remote.buildbuddy.io --host_platform=//:nix_deps_image_platform //...
```

So every time I bump/add a nix dependency, I push a new version of the image and update the value of `_NIX_DEPS_IMAGE_TAG`. That's the only change from the pov of bazel; as I said, I don't know what it does here (I mostly update the image when I bump the version of nixpkgs, and expect everything to be rebuilt anyway, so didn't pay attention)

core/nixpkgs.bzl
jcpetruzza: Possibly, and now that I think about it, it may end up removing spaces in the user string. Should probably use `"` instead, and manually escape `"`, `\` and `${`

core/nixpkgs.bzl
jcpetruzza: It's been a while, but I *think* the idea was that we want to have a `nixpkgs/default.nix` on every package, but then we need to prevent this interfering with user provided files, so by putting the user input under `nixpkgs/src`, we ensure there will be no conflict

dmadisetti: Hairiest merge res is around here. Didn't dig deep enough to figure out how to resolve- just took changes and hoped it worked

experimental/docker.bzl
```diff
@@ -0,0 +1,91 @@
...
+    # HACK! On bazel from nixpkgs, shebangs get mangled in things like
+    # @bazel_tools//tools/cpp:linux_cc_wrapper.sh.tpl, so that nix store
+    # paths end up referenced there.
+    # One needs to ensure those paths are available on the docker image
+    # as well, we can do that by including bazel
+    srcs = srcs + [bazel] if bazel else srcs
+
+    contents = []
+    for src in srcs:
+        path_to_default_nix = repository_ctx.path(src.relative("default.nix"))
```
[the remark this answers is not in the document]
dmadisetti: `touch default.nix` seemed to let things run for me. But should be more general (e.g. flakes, or nothing)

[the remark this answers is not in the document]
dmadisetti: Yeah, seems to want all deps to have a `default.nix`, even non-nix sources

```
Error in path: Not a regular file: /home/user/.cache/bazel/_bazel_user/8bbdcf70ee87482f767b06aac75f66d1/external/com_google_absl/absl/flags/default.nix
```

Could just check? I'll look into this a bit more

core/nixpkgs.bzl
aherrmann: I see, rules_nixpkgs [already creates a special sub-directory called `bazel-support`](https://github.com/tweag/rules_nixpkgs/blob/62d63740061dd94ecee3ec2c0cfc8a7ed409b98c/core/nixpkgs.bzl#L324), which houses the Nix output link used as GC-root.

So, user-code already has to avoid collision with that reserved name.

Would it be possible to generate the `nixpkgs/default.nix` under that sub-directory instead? Then this change wouldn't need to introduce the `nixpkgs/src` prefix.

aherrmann:
> I guess it depends on whether bazel considers the container-image of a (host) platform definition part of the build input or not, but I don't really know what it does here.

AFAIK it does, [see here](https://github.com/bazelbuild/bazel/blob/4e739d30431a31b5ab67187d29299809ac6c277a/third_party/remoteapis/build/bazel/remote/execution/v2/remote_execution.proto#L709-L712). So, I would expect this to invalidate the entire build. That also matches what we've seen in experiments in the past. 

> I mostly update the image when I bump the version of nixpkgs, and expect everything to be rebuilt anyway, so didn't pay attention

Fair enough. How problematic this is depends on the use-case. Indeed, bumping the nixpkgs revision will likely invalidate most targets in a project, but is also not expected to happen very frequently. Adding or tweaking individual packages, e.g. Nix imports of utilities developed separately, or third-party dependencies, OTOH should ideally not have that problem.

benradf:
Closing this as it has been superseded by #351. Will try and get that PR merged instead.

PR #217, opened 2022-04-12
add reference to examples in README

fricklerhandwerk:
help newcomers orient themselves how to actually use `rules_nixpkgs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ Links:
 [blog-bazel-nix]: https://www.tweag.io/posts/2018-03-15-bazel-nix.html
 [youtube-bazel-nix]: https://www.youtube.com/watch?v=hDdDUrty1Gw
 
+See [examples](/examples) for how to use `rules_nixpkgs` with different toolchains.
+
 ## Rules
 
 * [nixpkgs_git_repository](#nixpkgs_git_repository)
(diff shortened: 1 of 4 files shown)


steshaw (approved):
Looks good.

A nice follow-up might be to move the multiline strings containing strings to external `.md` files to make them easier to edit (with editor supported modes). If possible of course.

fricklerhandwerk:
It is possible and we had this until recently, but this made documentation rendering fairly complex as it required more build steps, and made the source harder to navigate. It's of course a trade-off. Right now it is not that much text and everything is self-contained in between the actual code, so it seems appropriate to me.

steshaw:
@fricklerhandwerk I'd like to take a look at the previous complexity. Can you point me at a commit?

fricklerhandwerk:
@steshaw #202 and [this specifically](https://github.com/tweag/rules_nixpkgs/pull/202/files#diff-f8ca881095fec0f1448b91970239419034e141839711f602e8cbc5131d1c061fL52-L115).

steshaw:
Thanks. I see! That does indeed look complicated. What's more, it doesn't help with editor tooling to have the markdown file in a template file `.md.tpl`. The complexity seems to be related to processing the `.tpl` file using AWK. I was thinking more of a `README.top.md` and `README.bottom.md` and then just concatenating those with the generated API docs with `cat`. That might be awkward though. It is a trade-off.

Issue #218, opened 2022-04-12
CXX_FLAGS should allow for overload


dmadisetti:
**Describe the bug**
Explicit definition of `-x c++` means I cannot use `cc_*` to compile cuda with clang...
https://github.com/tweag/rules_nixpkgs/blob/master/toolchains/cc/cc.nix#L164
https://github.com/tweag/rules_nixpkgs/blob/0d40e42ef8f2369638ee240cb81f1e6cab62a48d/toolchains/cc/cc.nix#L164

I don't see a way overloading this, and I don't think the explicit definition is needed - since compilers generally can figure that out by themselves. Currently just have a patch file that removes this line

**To Reproduce**
```
nixpkgs_cc_configure(                                                                                                                                                                                                             
    name = "nixpkgs_config_cc",                                                                                                                                                                                                   
    nix_file_content = """                                                                                                                                                                                                        
      with import <nixpkgs> { config = {allowUnfree = true; }; overlays = []; }; buildEnv {                                                                                                                                       
        name = "bazel-cc-toolchain";                                                                                                                                                                                              
        paths = [ clang_13 cudatoolkit_11 ];                                                                                                                                                                                      
        buildInputs = [  ];                                                                                                                                                                                                       
      }                                                                                                                                                                                                                           
    """,                                                                                                                                                                                                                          
    repository = "@nixpkgs",                                                                                                                                                                                                      
)
```
and try compiling some cuda (example: https://gist.github.com/dpiponi/1502434)
basically the same as `clang++ <cuda specific flags> -x c++ example.cu`, which will break since syntax unrecognized and c++ takes precedent.

**Environment**
NixOS unstable on master

dmadisetti:
I don't have an answer, maybe it's a clang change?
https://github.com/tweag/rules_nixpkgs/commit/6d6fc1b608a79ac2036688d640c3d13c68e158f2

I can investigate the earliest clang version that's compatible.

aherrmann:
See https://github.com/tweag/rules_nixpkgs/pull/128#issuecomment-729835453 and https://github.com/tweag/rules_nixpkgs/pull/128#issuecomment-732941362 for why the `-x c++` flag is necessary.

> I don't think the explicit definition is needed - since compilers generally can figure that out by themselves.

[...]

PR #219, opened 2022-04-12
cc: Remove explicit CXX_FLAG

dmadisetti:
Ran workflows on my branch and it seemed to work (except the hooks that need buildbuddy)
https://github.com/AnotherGroupChat/rules_nixpkgs/actions/runs/2157830144

Just seeing how #218 influences your tests. Let me know if there is a better way- but vanilla bazel does not add this flag

Opening as a draft as you all may have better ideas on how to implement this

### DIFF
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -161,7 +161,6 @@ in
         -fno-omit-frame-pointer
       )
       CXX_FLAGS=(
-        -x c++
         -std=c++0x
       )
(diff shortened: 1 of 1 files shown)


evertedsphere:
@dmadisetti is this still something you need? It seems like the upstream fix at https://github.com/bazelbuild/bazel/issues/15499 is also stalled.

evertedsphere:
Fixed in #298.

PR #220, opened 2022-04-13
Allow graceful failure when unable to capture external includes

dmadisetti:
Currently for something like clang-unwrapped nothing is grepped so this fails


dmadisetti:
Woops. I need to rebase this since there have been obvious changes. Sorry I haven't been up to date.

The command fails fails if it can't extract an include directory, causing the whole templating process to fail. I only came across this when I was exploring provisioning a tool-chain with `clang-unwrapped`. The `|| :` allows for for failure. I guess we could do `|| true` or explicitly check for the value before piping it.

This is no longer a blocker for me, but I think it is still a nice defensive check, since this shouldn't break.

dmadisetti:
I am closing this since I do not think I will have time to address this.

But just be aware, that this is a form of failure with cryptic error messages

avdv:
@dmadisetti  AFAIU, the problem is that if the `$cc -E -v ...` call somehow does not produce any output, or at least no `include <...>` section, then the `xargs -0 realpath -ms` will fail.

Instead of suppressing the error code, we could use the `-r` flag for xargs. Would this solve the problem you encountered?

PR #221, opened 2022-04-14
point examples to list of toolchains

fricklerhandwerk:
going there directly should be less confusing for new readers.


steshaw (approved):
Agreed

steshaw:
The changes look good but the build is failing because the generated READMEs need to be updated and checked-in.

```
README.md is not up to date.
Please update it using the following command:

bazel run //docs:update-README.md
```

PR #227, opened 2022-05-17
Use `nix-support/cc-{c,ld}flags` files to configure compiler flags

### LINKED ISSUE #224: Support codesigning on MacOs
On M1 MacOS Monterey it seems to be mandatory for binaries to be codesigned, even binaries that are self-built. The default XCode toolchain handles this transparently. However, a Nix provided Bazel CC toolchain needs to take extra care to support this.

Configure the Nix provided CC toolchain on Darwin to use codesign. This has been implemented in Daml [here](https://github.com/digital-asset/daml/blob/fb3ecdc2c2354c358f154ed1cab22c68a722ae9d/nix/tools/bazel-cc-toolchain/default.nix#L20-L30) and needs to be upstreamed.

### LINKED ISSUE #225: Use the nix-support/cc-flags file to configure compiler flags
Currently we configure default compiler flags in the Nix provided CC toolchain by creating a wrapper script around the compiler binary. However, nixpkgs already has a feature to configure such flags through the `cc-flags` and `ld-flags` file in the `nix-support` directory.

An issue with the wrapper approach is that the flags are passed indiscriminately of how the compiler is invoked. In certain invocations this can cause incompatibilities. For example, on MacOS Monterey on M1 with an LLVM toolchain it was found that calling the compiler with `-v` and any `-l` or `-isystem` flags caused errors. See https://github.com/digital-asset/daml/pull/13401/commits/fb3ecdc2c2354c358f154ed1cab22c68a722ae9d for details. 

Configure any additional flags to the CC toolchain using the `cc-flags` or `ld-flags` files. An implementation that can be upstreamed exists here https://github.com/digital-asset/daml/pull/13401/commits/fb3ecdc2c2354c358f154ed1cab22c68a722ae9d

**Additional context**
This was discovered while working on M1 support on [Daml here](https://github.com/digital-asset/daml/pull/13401).

avdv:
# Purpose

This improves support for using rules_nixpkgs on Darwin where it is required to sign binaries.

Fixes #225 and fixes #224.

### DIFF
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -22,8 +22,8 @@ let
         echo "-F${CoreServices}/Library/Frameworks" >> $out/nix-support/cc-cflags
         echo "-F${Security}/Library/Frameworks" >> $out/nix-support/cc-cflags
         echo "-F${Foundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
-        echo "-L${pkgs.libiconv}/lib" >> $out/nix-support/cc-ldflags
-        echo "-L${pkgs.darwin.libobjc}/lib" >> $out/nix-support/cc-ldflags
+        echo "-L${pkgs.libiconv}/lib" >> $out/nix-support/cc-cflags
+        echo "-L${pkgs.darwin.libobjc}/lib" >> $out/nix-support/cc-cflags
       '';
     };
   cc =
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -34,25 +34,16 @@ let
     else
       pkgs.buildEnv (
         let
-          paths =
-            if pkgs.stdenv.isDarwin then
-              {
-                inherit (darwinCC) cc bintools;
-              }
-            else
-              {
-                cc = pkgs.stdenv.cc;
-                bintools = pkgs.binutils;
-              };
+          cc = if pkgs.stdenv.isDarwin then darwinCC else pkgs.stdenv.cc;
         in
         {
           name = "bazel-nixpkgs-cc";
           # XXX: `gcov` is missing in `/bin`.
           #   It exists in `stdenv.cc.cc` but that collides with `stdenv.cc`.
-          paths = [ paths.cc paths.bintools ];
+          paths = [ cc cc.bintools ];
           pathsToLink = [ "/bin" ];
           passthru = {
-            isClang = paths.cc.isClang;
+            isClang = cc.isClang;
           };
         }
       )
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -17,11 +17,14 @@ let
       bintools = cc.bintools;
       extraBuildCommands = with pkgs.darwin.apple_sdk.frameworks; ''
         echo "-Wno-unused-command-line-argument" >> $out/nix-support/cc-cflags
-        echo "-isystem ${pkgs.llvmPackages.libcxx}/include/c++/v1" >> $out/nix-support/cc-cflags
+        echo "-isystem ${pkgs.llvmPackages.libcxx.dev}/include/c++/v1" >> $out/nix-support/cc-cflags
+        echo "-isystem ${pkgs.llvmPackages.clang-unwrapped.lib}/lib/clang/${cc.version}/include" >> $out/nix-support/cc-cflags
         echo "-F${CoreFoundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
         echo "-F${CoreServices}/Library/Frameworks" >> $out/nix-support/cc-cflags
         echo "-F${Security}/Library/Frameworks" >> $out/nix-support/cc-cflags
         echo "-F${Foundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-L${pkgs.llvmPackages.libcxx}/lib" >> $out/nix-support/cc-cflags
+        echo "-L${pkgs.llvmPackages.libcxxabi}/lib" >> $out/nix-support/cc-cflags
         echo "-L${pkgs.libiconv}/lib" >> $out/nix-support/cc-cflags
         echo "-L${pkgs.darwin.libobjc}/lib" >> $out/nix-support/cc-cflags
       '';
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -9,12 +9,12 @@ in
 }:
 
 let
-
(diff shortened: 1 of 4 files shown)


toolchains/cc/cc.nix
```diff
@@ -37,20 +37,19 @@ let
...
               {
-                cc = (pkgs.overrideCC pkgs.stdenv darwinCC).cc;
-                binutils = pkgs.darwin.binutils;
+                inherit (darwinCC) cc bintools;
               }
             else
               {
                 cc = pkgs.stdenv.cc;
-                binutils = pkgs.binutils;
+                bintools = pkgs.binutils;
```
avdv, line 45: should we just use `pkgs.stdenv.cc.bintools` here for consistency too?

avdv: They seem to be the same by default:
```
# MacOS
nix-repl> darwin.binutils
_derivation /nix/store/rvjqxkb0df4nlv4lvqljshf7ac3y54c8-cctools-binutils-darwin-wrapper-949.0.1.drv_

nix-repl> stdenv.cc.bintools
_derivation /nix/store/rvjqxkb0df4nlv4lvqljshf7ac3y54c8-cctools-binutils-darwin-wrapper-949.0.1.drv_
```
```
# Linux
nix-repl> stdenv.cc.bintools
«derivation /nix/store/fwywivmkyn0mg7qbsy161zzqfv9kq415-binutils-wrapper-2.35.2.drv»

nix-repl> binutils
«derivation /nix/store/fwywivmkyn0mg7qbsy161zzqfv9kq415-binutils-wrapper-2.35.2.drv»
```
I was thinking that if somebody alters the `stdenv` of the packages set we get passed in, it would be counter-intuitive if we use the `cc` from `stdenv` (which was overridden) but insist on using the plain `binutils` package.

aherrmann: Thanks for checking, yes, in that case being consistent seems like the better approach.

toolchains/cc/cc.nix
```diff
@@ -12,32 +12,20 @@ let
...
+      bintools = cc.bintools;
+      extraBuildCommands = with pkgs.darwin.apple_sdk.frameworks; ''
+        echo "-Wno-unused-command-line-argument" >> $out/nix-support/cc-cflags
+        echo "-isystem ${pkgs.llvmPackages.libcxx}/include/c++/v1" >> $out/nix-support/cc-cflags
+        echo "-F${CoreFoundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-F${CoreServices}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-F${Security}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-F${Foundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-L${pkgs.libiconv}/lib" >> $out/nix-support/cc-ldflags
+        echo "-L${pkgs.darwin.libobjc}/lib" >> $out/nix-support/cc-ldflags
```
aherrmann, line 26:
```suggestion
        echo "-L${pkgs.libiconv}/lib" >> $out/nix-support/cc-cflags
        echo "-L${pkgs.darwin.libobjc}/lib" >> $out/nix-support/cc-cflags
```
I think these need to go into `cc-cflags` as well instead of `cc-ldflags`. That's also where they are added in the [daml changes](https://github.com/digital-asset/daml/commit/fb3ecdc2c2354c358f154ed1cab22c68a722ae9d#diff-e6d8f61a3aac62da04942e4ffdac19c808513510784fa31f63c8d061ad52d104R46-R47). The reason is that Bazel will invoke the cc-wrapper for linking as well. So, adding these flags only to `cc-ldflags` may not enable them in all needed use-cases.

toolchains/cc/cc.nix
```diff
@@ -12,32 +12,20 @@ let
...
+      bintools = cc.bintools;
+      extraBuildCommands = with pkgs.darwin.apple_sdk.frameworks; ''
+        echo "-Wno-unused-command-line-argument" >> $out/nix-support/cc-cflags
+        echo "-isystem ${pkgs.llvmPackages.libcxx}/include/c++/v1" >> $out/nix-support/cc-cflags
+        echo "-F${CoreFoundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-F${CoreServices}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-F${Security}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-F${Foundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
+        echo "-L${pkgs.libiconv}/lib" >> $out/nix-support/cc-cflags
+        echo "-L${pkgs.darwin.libobjc}/lib" >> $out/nix-support/cc-cflags
```
aherrmann, line 26: Comparing to [daml](https://github.com/digital-asset/daml/commit/fb3ecdc2c2354c358f154ed1cab22c68a722ae9d#diff-e6d8f61a3aac62da04942e4ffdac19c808513510784fa31f63c8d061ad52d104R36-R47) I see some paths missing
```suggestion
        echo "-Wno-unused-command-line-argument" >> $out/nix-support/cc-cflags
        echo "-isystem ${pkgs.llvmPackages.libcxx.dev}/include/c++/v1" >> $out/nix-support/cc-cflags
        echo "-isystem ${pkgs.llvmPackages.clang-unwrapped.lib}/lib/clang/${cc.version}/include" >> $out/nix-support/cc-cflags
        echo "-F${CoreFoundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
        echo "-F${CoreServices}/Library/Frameworks" >> $out/nix-support/cc-cflags
        echo "-F${Security}/Library/Frameworks" >> $out/nix-support/cc-cflags
        echo "-F${Foundation}/Library/Frameworks" >> $out/nix-support/cc-cflags
        echo "-L${pkgs.llvmPackages.libcxx}/lib" >> $out/nix-support/cc-cflags
        echo "-L${pkgs.llvmPackages.libcxxabi}/lib" >> $out/nix-support/cc-cflags
        echo "-L${pkgs.libiconv}/lib" >> $out/nix-support/cc-cflags
        echo "-L${pkgs.darwin.libobjc}/lib" >> $out/nix-support/cc-cflags
```

aherrmann (approved):
Thanks a lot for porting this over and getting it in shape! Assuming this also works on M1, I think it's good to merge!

avdv:
> Assuming this also works on M1, I think it's good to merge!

PR #230, opened 2022-05-27
Fix error when using an unreleased Bazel version

### LINKED ISSUE #229: Problem when running with a non-release Bazel version
After [rules_haskell](https://github.com/tweag/rules_haskell) updated their `rules_nixpkgs` version in commit https://github.com/tweag/rules_haskell/commit/bfcec89006d4e54b523847ccddccfba36c3a12fc, their tests with Bazel@Head started to fail ([example](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2480#ccc56079-febc-47c3-bbfd-e1d05c868520)). I think the reason is that [this check](https://github.com/tweag/rules_nixpkgs/blob/master/toolchains/python/python.bzl#L59) that was added in the new version and it does not take into account using non-release Bazel binaries.

avdv:
Fixes #229

### DIFF
--- a/core/.bazelrc
+++ b/core/.bazelrc
@@ -16,6 +16,7 @@ build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_hostjdk8
 # ----------------
 common:ci --color=no
 build:ci --verbose_failures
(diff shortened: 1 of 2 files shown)


avdv:
superseded by #231

PR #233, opened 2022-05-31
Run all tests and accumulate failures on CI

avdv:
After splitting the repository in multiple modules (in https://github.com/tweag/rules_nixpkgs/pull/193) tests are only run until the first error.

In case Bazel crashes due to build stream errors, this does not indicate a problem at our end. However, we should ensure that all tests passed to make an informed decision whether we can merge a PR even if the CI is not green.

----

Maybe, (just maybe) this also seems to have fixed the BuildBuddy crash issue -- all tests ran fine and there are now 10 tests reported at https://app.buildbuddy.io/history/repo/tweag/rules_nixpkgs

### DIFF
--- a/.github/build-and-test
+++ b/.github/build-and-test
@@ -1,0 +1,47 @@
+#! /usr/bin/env nix-shell
+#! nix-shell ../shell.nix -i bash
+
+set -euo pipefail
+
+declare SCRIPT_DIR
+
+SCRIPT_DIR="$( realpath "${BASH_SOURCE[0]}" )"
+SCRIPT_DIR="${SCRIPT_DIR%/*}"
+
+cd "$( dirname "$SCRIPT_DIR" )"
+
+declare -i FAILED=0
+declare -a FAILURES=( )
+
+function on_exit() {
+    echo
+    if [[ "$FAILED" -gt 0 ]]; then
+        echo "error: running tests in ${FAILURES[@]} was not successful"
+        exit 1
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yaml
```diff
@@ -25,25 +25,7 @@ jobs:
...
-              toolchains/rust
-            )
-            for dir in "${dirs[@]}"
-            do
-              pushd $dir
-                bazel test //...
-              popd
-            done
-          '
+          nix-shell --pure --keep GITHUB_REPOSITORY --run 'bash .github/build-and-test'
```
aherrmann, line 28: `--keep GITHUB_REPOSITORY` out of curiosity, where do we need that?

.github/build-and-test
```diff
@@ -0,0 +1,48 @@
...
+    toolchains/rust
+)
+for dir in "${dirs[@]}"
+do
+    pushd $dir
+    if ! bazel test //... ; then
+        let ++FAILED
+        FAILURES+=( "$dir" )
+    fi
+    bazel shutdown
```
aherrmann, line 45: This may be the thing that fixed the BuildBuddy issue, IIRC it was suggested in the past. Is there any other motivation for this step?

aherrmann (approved):
Thank you!

.github/workflows/workflow.yaml
aherrmann: Ah, thanks for clarifying!

PR #234, opened 2022-06-01
Avoid error when include dirs list is empty

avdv:
Originally raised by @dmadisetti in #220.

This PR just skips running the `xargs` command if there are no arguments read from stdin.


dmadisetti:
LGTM!

aherrmann (approved):
Thanks!

PR #237, opened 2022-06-10
Avoid linker warning about non-existent directory

avdv:
The `wrapCCWith` function adds the `$cc/lib` directory to the library search path (`-L`) using the
`nix-support/cc-ldflags` file.

Usually, for `stdenv.cc` this directory does not exist and `ld` will warn "directory not found for option `-L...`.

More details and an alternative approach fixing this is described here: https://github.com/tweag/rules_haskell/pull/1750#issuecomment-1151228730


aherrmann (approved):
Thank you! Yes, that's nicer than the `mkdir` approach.

PR #240, opened 2022-06-28
Add support for cross cc compilers

layus:
In the current state, cc.nix is not compatible with crossPkgs and cross compilers.

This PR makes some progress by using buildPackages instead of pkgs where needed (everywhere, except stdenv.cc).

To get a cross toolchain one still needs to provide a nixpkgs with an hard-coded cross config. Alternatively, an expression that provides a cross compiler should also work without modifying the nixpkgs argument to nixpkgs_cc_configure.

But providing such an expression is difficult because the wrapper (the `buildEnv` call) needs to be duplicated. Any reason why we do not always apply it ? Worst case the user expression needs to be coaxed a bit to provide binutils, isClang and targetPrefix. Not really hard.

A tricky one is also wrapCCWith. Not sure how it works in mixed setups.

Finally, I could use some help writing tests in this new repo layout.


avdv:
Nice! You beat me to it... :smile: 

Note, that this is not only useful for cross-compilation, but also when using a static CC based on musl for example (think `pkgs.pkgsStatic.clangStdenv.cc` for example).

I'll have a deeper look when I found some time later...

aherrmann:
Nice! Thank you for looking into this!

> But providing such an expression is difficult because the wrapper (the buildEnv call) needs to be duplicated. Any reason why we do not always apply it?

Mainly, when we created this, we didn't know how much liberty users would need to override the details of the toolchain, e.g. see [here](https://github.com/digital-asset/daml/blob/d0058dcff357e36bf70f3dcf130f0cd73f056599/nix/tools/bazel-cc-toolchain/default.nix). However, looking at the shape this is taking, and how things like `binutils` can be bundled with `cc`, it may well be fine to hard-code the `buildEnv` and only take the `cc` as a parameter.

> Finally, I could use some help writing tests in this new repo layout.

Where are you encountering difficulties? The tests for the cc-toolchain have been moved [here](https://github.com/tweag/rules_nixpkgs/blob/master/toolchains/cc/tests/BUILD.bazel).

aherrmann:
@layus What's the status on this one? What do you need to push it over the finish line?

avdv (approved):
LGTM :+1:

benradf (approved):

PR #251, opened 2022-07-14
[PoC] Package up `binary` targets & dependencies in a nix package or docker container

ryanrasti:
Hello! I stumbled on nix and this repo a few days ago and really love the idea. Thanks so much for creating it. I can see so much potential here.

Anyways, I am trying to create a rule to package up a binary target into a docker container (with all of its nix dependencies) with bazel. e.g., in a `rule.bzl` file I have a command to generate a docker tar file:
```python
load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_package")

def nix_image(name, target, data, visibility=None):
  # TODO: still need to get runfiles in!
  native.genrule(
    name = 'gen_image_%s' % name,
    outs = [name + ".tar.gz"],
    cmd = r'''
# TODO: hacky -- want to take in the transitive deps explicitly!

# find the *derivers* for the outputs paths:
f=$$(find | grep bazel-support/nix-out-link.drv)
echo "Found derivers: $$f"
lines=$$(echo $$f | xargs realpath | paste -sd "," -)
echo nix-build $$(realpath $(location //nix:docker.nix)) --arg lines \"$$lines\" >&2
build_out=$$(nix-build $$(realpath $(location //nix:docker.nix)) --arg lines \"$$lines\")
echo out: $$build_out >&2
ln -s $$build_out $(OUTS)
''',
    tools=["//nix:docker.nix"] + data + [target],
    visibility = visibility,
  )
```

Notice the above (hackily) looks for any `bazel-support/nix-out-link.drv` in the build context and then passes them all to this derivation (in `//nix:docker.nix`):

```nix
{ pkgs ? import <nixpkgs> {}, lines ? "" }:
let
  x = break;
  imports = map (l: import l) (pkgs.lib.strings.splitString "," lines);
in
pkgs.dockerTools.buildImage {
    name = "test-nix";
    tag = "latest";

    copyToRoot = pkgs.buildEnv {
      name = "image-root";
      paths = [pkgs.coreutils pkgs.bashInteractive] ++ imports;
      pathsToLink = [ "/bin" "/lib" ];
    };
}
```

But to get the `.drv` files which are needed as inputs (I couldn't figure out how to `import` the built derivations as inputs), I had to create the changes in this PR.

I am wondering if there is a way to build an image like this without having to modify `rules_nixpkgs`? If not, what'd be the best way to support it?

### DIFF
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -187,11 +187,13 @@ def _nixpkgs_package_impl(repository_ctx):
         # a file named `bazel-support` in its root.
         # A `bazel clean` deletes the symlink and thus nix is free to garbage collect
         # the store path.
-        "--out-link",
-        out_link,
     ])
 
-    expr_args.extend([
+    inst_args = expr_args + ["--add-root", out_link + '.drv']
+    expr_args += ["--out-link", out_link]
+
+    end_args = []
+    end_args.extend([
         expand_location(
             repository_ctx = repository_ctx,
             string = opt,
@@ -200,6 +202,7 @@ def _nixpkgs_package_impl(repository_ctx):
         )
         for opt in repository_ctx.attr.nixopts
     ])
+
 
     for repo in repositories.keys():
         path = str(repository_ctx.path(repo).dirname) + "/nix-file-deps"
@@ -220,7 +223,7 @@ def _nixpkgs_package_impl(repository_ctx):
         fail(strFailureImplicitNixpkgs)
 
     for dir in nix_path:
-        expr_args.extend(["-I", dir])
+        end_args.extend(["-I", dir])
 
     if not_supported and fail_not_supported:
         fail("Platform is not supported: nix-build not found in PATH. See attribute fail_not_supported if you don't want to use Nix.")
@@ -232,7 +235,7 @@ def _nixpkgs_package_impl(repository_ctx):
             "nix-build",
             extra_msg = "See: https://nixos.org/nix/",
         )
-        nix_build = [nix_build_path] + expr_args
+        nix_build = [nix_build_path] + expr_args + end_args
 
         # Large enough integer that Bazel can still parse. We don't have
         # access to MAX_INT and 0 is not a valid timeout so this is as good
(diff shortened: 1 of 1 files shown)


ryanrasti:
A very common case for me is that I want to rely on some binary being in PATH so that when I depend on a nix target in bazel I can easily invoke it via a subprocess. Do you currently have a recommendation for this?

Another solution I used is to have a wrapper script for my py_runtime that looks for all of the nix packages in runfiles and then dynamically forms PATH from all of their `bin` directories. But I am not sure if that is generally supposed to work.

aherrmann:
> A very common case for me is that I want to rely on some binary being in PATH so that when I depend on a nix target in bazel I can easily invoke it via a subprocess. Do you currently have a recommendation for this?

I'm not sure how this relates to Docker, could you expand on that?

Nonetheless, coming from Nix, this is understandable, as Nix manages `PATH` automatically in such a way. Bazel, OTOH, doesn't do that. Instead, it's usually on the rule author to manually extend `PATH` as needed. That said, a recent addition to `rules_sh`, namely [`sh_binaries`](https://github.com/tweag/rules_sh/blob/06170f1fbff23f900bc04dfbdb75ed1acf2b8c42/sh/sh.bzl#L162), may be of interest to you. With that you could create a Nix import of the following form:
```
# WORKSPACE
nixpkgs_package(
    name = "hello",
    attribute_path = "hello",
    build_file_content = """\
load("@rules_sh//sh:sh.bzl", "sh_binaries")
sh_binaries(
    name = "hello",
    srcs = ["bin/hello"],
    visibility = ["//visibility:public"],
)
""",
)
```
And then use it as follows
```
genrule(
    name = "some-genrule",
    toolchains = ["@hello"],
    cmd = """\
PATH="$(_HELLO_PATH):$$PATH"
hello >$@
""",
    outs = ["output"],
)
```

benradf:
Closing as it's been nearly a year with no changes now. Feel free to re-open if this is something you're still interested in getting merged @ryanrasti.

PR #257, opened 2022-08-05
Update bzlmod dependencies and enhance separation of components

avdv:
This PR takes bzlmod migration a step further and tries to lay ground for future work in that area.

Currently, there are a few components / modules which were all depending on each other:

* `core`
* toolchains:
    * cc
    * go
    * java
    * posix
    * rust
    * python
* nixpkgs (the *everything* component)

It should be possible to require a specific toolchain with only pulling in the required dependencies.

This PR adds a `toolchains` parameter to the `rules_nixpkgs_dependencies` function signalling which toolchain is required. 

Interestingly, every component depends on java, posix and the cc toolchain (even the core). This is because of the stardoc dependency which consists of a `java_binary`...

IMO, it would be a good idea to separate the docs generation into another component, in order to get rid of the excess dependencies. WDYT?

### DIFF
--- a/core/MODULE.bazel
+++ b/core/MODULE.bazel
@@ -4,4 +4,6 @@ module(
 )
 
 bazel_dep(name = "platforms", version = "0.0.4")
-bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "bazel_skylib", version = "1.0.3", repo_name = "bazel_skylib")
+bazel_dep(name = "rules_sh", version = "0.3.0")
+bazel_dep(name = "stardoc", version = "0.5.1", repo_name = "io_bazel_stardoc")
--- a/nixpkgs/repositories.bzl
+++ b/nixpkgs/repositories.bzl
@@ -2,11 +2,12 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
-def rules_nixpkgs_dependencies(rules_nixpkgs_name = "io_tweag_rules_nixpkgs"):
+def rules_nixpkgs_dependencies(rules_nixpkgs_name = "io_tweag_rules_nixpkgs", toolchains = None):
     """Load repositories required by rules_nixpkgs.
 
     Args:
         rules_nixpkgs_name: name under which this repository is known in your workspace
+        toolchains:         list of toolchains to load, ['cc', 'java'], load all toolchains by default
     """
     maybe(
         http_archive,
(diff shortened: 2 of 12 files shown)


aherrmann:
> IMO, it would be a good idea to separate the docs generation into another component, in order to get rid of the excess dependencies. WDYT?

That sounds like a reasonable approach to me, if feasible. FWIW modules also introduce a notion of dev-dependencies, so, perhaps in that case the components needed for Stardoc could simply be marked as such and wouldn't be needed for regular use. cc @fricklerhandwerk any thoughts? I think you're more familiar with this part than me.

core/MODULE.bazel
```diff
@@ -4,4 +4,6 @@ module(
 )
 
 bazel_dep(name = "platforms", version = "0.0.4")
-bazel_dep(name = "bazel_skylib", version = "1.0.3")
+bazel_dep(name = "bazel_skylib", version = "1.0.3", repo_name = "bazel_skylib")
+bazel_dep(name = "rules_sh", version = "0.3.0")
+bazel_dep(name = "stardoc", version = "0.5.1", repo_name = "io_bazel_stardoc")
```
aherrmann, line 9: Could this be a dev-dependency?

nixpkgs/repositories.bzl
```diff
@@ -2,11 +2,12 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
-def rules_nixpkgs_dependencies(rules_nixpkgs_name = "io_tweag_rules_nixpkgs"):
+def rules_nixpkgs_dependencies(rules_nixpkgs_name = "io_tweag_rules_nixpkgs", toolchains = None):
     """Load repositories required by rules_nixpkgs.
 
     Args:
         rules_nixpkgs_name: name under which this repository is known in your workspace
+        toolchains:         list of toolchains to load, ['cc', 'java'], load all toolchains by default
```
aherrmann, line 10:
```
e.g `['cc', 'java']`
```

nixpkgs/repositories.bzl
```diff
@@ -50,14 +51,10 @@ def rules_nixpkgs_dependencies(rules_nixpkgs_name = "io_tweag_rules_nixpkgs"):
...
-    for name, prefix in [
-        ("rules_nixpkgs_core", "core"),
-        ("rules_nixpkgs_cc", "toolchains/cc"),
-        ("rules_nixpkgs_java", "toolchains/java"),
-        ("rules_nixpkgs_python", "toolchains/python"),
-        ("rules_nixpkgs_go", "toolchains/go"),
-        ("rules_nixpkgs_rust", "toolchains/rust"),
-        ("rules_nixpkgs_posix", "toolchains/posix"),
+    for name, prefix in [("rules_nixpkgs_core", "core")] + [
+        ("rules_nixpkgs_" + toolchain, "toolchains/" + toolchain)
```
aherrmann, line 55: Perhaps needs a check to guard against non-existent toolchains. Otherwise, users could see very confusing error messages on typos.

aherrmann (commented):
Thank you for looking into this! Looks good, only a question about dev-dependencies.

core/MODULE.bazel
avdv: Yes, changed.

nixpkgs/repositories.bzl
avdv: Done.

nixpkgs/repositories.bzl
avdv: Good catch! I added a check to error out if an unknown toolchain is passed in.

aherrmann (approved):
Thanks for the updates, that looks good!

> IMO, it would be a good idea to separate the docs generation into another component, in order to get rid of the excess dependencies. WDYT?

Is this something you want to tackle in a follow-up PR?

fricklerhandwerk (approved):

avdv:
> Is this something you want to tackle in a follow-up PR?

Yes, I created an issue to track this: https://github.com/tweag/rules_nixpkgs/issues/264

Issue #256, opened 2022-08-02
Slow nixpkgs_package rule on macOS


lamcw:
**Describe the bug**
`nixpkgs_package` runs slow (~5 minutes) trying to copy all `nix_file_deps` to external repository root.

I noticed slow `nixpkgs_package` fetches, and upon investigation, I found out that

1. `nixpkgs_package` kept being restarted, possibly due to missing dependencies for the rule itself, or something related to https://github.com/bazelbuild/bazel/issues/10515
2. copying all `nix_file_deps` using [`cp`](https://github.com/tweag/rules_nixpkgs/blob/9f08fb2322050991dead17c8d10d453650cf92b7/core/util.bzl#L8) takes most of the time in `_nixpkgs_package_impl`. See screenshot of bazel profile below -- each `cp` call takes 50ms to 60ms to run, multiply that by 4200 invocations it can delay the build by up to 5 minutes.
    ![image](https://user-images.githubusercontent.com/25677733/182270172-4322b5a5-5edd-4684-a368-5dea02fe112b.png)
    _time to run one single `cp` call_

    ![image](https://user-images.githubusercontent.com/25677733/182271164-b35b1b9a-3aae-4780-82f6-10be40555285.png)
    _number of `cp` invocations in a `bazel build` command_

    And found out that by removing the following `chmod` call in the `cp` function improves the performance significantly (5mins -> 1min) so it seems like this is the bottleneck for the `nixpkgs_package` repository rule
    https://github.com/tweag/rules_nixpkgs/blob/9f08fb2322050991dead17c8d10d453650cf92b7/core/util.bzl#L41-L46
    My guess is that syscalls (`stat`s) on macOS is slow and therefore creates a bottleneck for the WORKSPACE analysis.

So wanted to raise the issue for awareness, and also this might be good information for others. Is there any workarounds to _not_ use `chmod` while retaining file permissions when copying files?

**To Reproduce**
In `WORKSPACE`, define the following
```starlark
# big list of nix files that `nixpkgs.nix` depends on
# for me 70 files is enough to hold up the build for 5 minutes
nix_files = [...]

nixpkgs_package(
    name = "python3",
    attribute_path = "python39",
    build_file_content = """
package(default_visibility = ["//visibility:public"])
filegroup(
    name = "interpreter",
    srcs = ["bin/python"],
)
    """,
    nix_file = ":nixpkgs.nix",
    nix_file_deps = nix_files,
    repository = "@nixpkgs",
)
[...]

PR #258, opened 2022-08-11
Add copy_many function to address performance issue

benradf:
This is an attempt to improve the MacOS performance issue described in #256 by adding a `cp_many` function to `util.bzl`. Instead of writing each file with `repository_ctx.file` and invoking `chmod` to fix the permissions, `cp_many` uses `tar` from an external shell process.

My hope is this will be more efficient than spawning a new `chmod` process for every file. However I do not have a Mac to test on. On Linux I attempted to reproduce the slowness by including 4096 random 16K files in `nix_file_deps`. For the original code that took 7m59s to run, and with this change it took 7m27s. So not much of an improvement, but maybe the speed up is greater on MacOS?


core/util.bzl
```diff
@@ -47,6 +47,70 @@ def cp(repository_ctx, src, dest = None):
...
+      srcs: A list of source files to copy.
+
+    Returns:
+      A list of destination paths for the files copied into the repository.
+    """
+    tar_list = []
+    repository_files = []
+
+    # Tar will run with outputBase as its working directory.
+    repository_dir = repository_ctx.execute(["pwd"]).stdout.strip("\n")
```
aherrmann, line 68: [`repository_ctx.path`](https://bazel.build/rules/lib/repository_ctx#path) resolves relative to the current workspace directory, so invoking `pwd` should not be needed.

core/util.bzl
```diff
@@ -47,6 +47,70 @@ def cp(repository_ctx, src, dest = None):
...
+    output_base = repository_dir.removesuffix("external/" + repository_ctx.name)
+
+    # Fallback to cp if outputBase cannot be determined.
+    if output_base == repository_dir:
+        return [cp(repository_ctx, src) for src in srcs]
+
+    for src in srcs:
+        abs_path = str(repository_ctx.path(src))
+        rel_path = abs_path.removeprefix(output_base)
```
aherrmann, line 77: [Bazel Skylib](https://github.com/bazelbuild/bazel-skylib/blob/872e9b06e18ae8ba2897cb65b9aaa172aa6279f3/lib/paths.bzl#L15) offers a file filepath mangling functions that could be better suited here.

core/util.bzl
```diff
@@ -47,6 +47,70 @@ def cp(repository_ctx, src, dest = None):
...
+        if rel_path != abs_path:
+            # This file can be copied with tar.
+            tar_list.append((src, rel_path))
+        else:
+            # If file is not under outputBase, fallback to cp.
+            # This can happen if Bazel is running in Bazel and
+            # the file was generated by the outer Bazel.
+            repository_files.append(cp(repository_ctx, src))
```
aherrmann, line 85: Why can't `tar` be applied to such paths?

core/util.bzl
```diff
@@ -47,6 +47,70 @@ def cp(repository_ctx, src, dest = None):
...
+
+    Returns:
+      A list of destination paths for the files copied into the repository.
+    """
+    tar_list = []
+    repository_files = []
+
+    # Tar will run with outputBase as its working directory.
+    repository_dir = repository_ctx.execute(["pwd"]).stdout.strip("\n")
+    output_base = repository_dir.removesuffix("external/" + repository_ctx.name)
```
aherrmann, line 69: Why does `tar` need to run in `outputBase`?

core/util.bzl
```diff
@@ -47,6 +47,70 @@ def cp(repository_ctx, src, dest = None):
...
+            # the file was generated by the outer Bazel.
+            repository_files.append(cp(repository_ctx, src))
+
+    # Write a list of paths to be included in the tar archive.
+    tar_list_content = "\n".join([p for (_, p) in tar_list])
+    tar_list_filename = "tarlist{}".format(hash(tar_list_content))
+    repository_ctx.file(tar_list_filename, tar_list_content)
+
+    # Invoke external shell process to copy files with tar.
+    tar_result = repository_ctx.execute(["sh", "-c", """
```
aherrmann, line 93: [`execute`](https://bazel.build/rules/lib/repository_ctx#execute) has a `working_directory` parameter, so the `cd` shouldn't be necessary.

aherrmann (commented):
@benradf Thanks for opening this! It would be great to know if this provides a substantial speedup over the per-file copy on MacOS. @lamcw is this something you could try out?

core/nixpkgs.bzl
```diff
@@ -88,10 +89,9 @@ def _nixpkgs_local_repository_impl(repository_ctx):
     else:
         target = cp(repository_ctx, repository_ctx.attr.nix_file)
 
-    repository_files = [target]
-    for dep in repository_ctx.attr.nix_file_deps:
-        dest = cp(repository_ctx, dep)
-        repository_files.append(dest)
+
+    repository_files = cp_many(repository_ctx, repository_ctx.attr.nix_file_deps)
```
aherrmann, line 93: This will probably require a [similar platform check to `cp`](https://github.com/tweag/rules_nixpkgs/blob/9f08fb2322050991dead17c8d10d453650cf92b7/core/util.bzl#L40) to not break Windows users.

core/util.bzl
benradf: Good suggestion - I'll take a look :+1:

core/util.bzl
benradf: Nice, will change.

core/util.bzl
benradf: `repository_dir = repository_ctx.path(".")` seems to do the trick.

core/util.bzl
aherrmann: I see, makes sense. I guess with GNU tar one could also use `--transform`, but I'm not sure if BSD tar has such an option.

core/util.bzl
aherrmann: Thanks for explaining, makes sense. Same thought about `--transform` [as here](https://via.hypothes.is/https://www.aip.org/history-programs/niels-bohr-library/oral-histories/5020-5). But, may not be available on BSD...

lamcw:
Thanks for the fix @benradf ! Just tested on my Mac and it seems to have around the same amount of improvement as my `stat` solution (also improves the build from 5 mins to 1 min).

benradf:
Thanks for checking this on Mac @lamcw. It's good to hear the performance improvement there is greater than on Linux.

Though since it's about the same as your `stat` solution I think I'll park this PR for now. Seems to better to avoid shelling out to `tar` if it doesn't provide a significant improvement.

benradf:
Going with #259 as a solution instead.

PR #259, opened 2022-08-12
Use stat to retain file permissions when copying files

### LINKED ISSUE #256: Slow nixpkgs_package rule on macOS

lamcw:
With stat instead of chmod, this should improve the performance of copying many files with `cp` as chmod runs multiple syscalls when changing permissions.

Would be great if I can get your help to test these changes on Linux as well!

This is an alternative solution to #258.

Closes #256.

### DIFF
--- a/core/util.bzl
+++ b/core/util.bzl
@@ -5,6 +5,25 @@ load("@bazel_skylib//lib:versions.bzl", "versions")
 def is_supported_platform(repository_ctx):
     return repository_ctx.which("nix-build") != None
 
+def _is_executable(repository_ctx, path):
+    stat_exe = repository_ctx.which("stat")
+    if stat_exe == None:
+        return False
+
+    # A hack to detect if stat in Nix shell is BSD stat as BSD stat does not
+    # support --version flag
+    is_bsd_stat = repository_ctx.execute([stat_exe, "--version"]).return_code != 0
+    if is_bsd_stat:
+        stat_args = ["-f", "%Lp", path]
+    else:
+        stat_args = ["-c", "%a", path]
+
+    arguments = [stat_exe] + stat_args
+    exec_result = repository_ctx.execute(arguments)
+    stdout = exec_result.stdout.strip()
+    mode = int(stdout, 8)
+    return mode & 0o100 != 0
+
 def cp(repository_ctx, src, dest = None):
     """Copy the given file into the external repository root.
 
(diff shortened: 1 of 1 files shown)


benradf:
I gave it a try on Linux and it seems to work fine :+1: 

One thing that I think is okay but wanted to explicitly call out – the file copies will not retain exactly the same permissions as the originals now that we're no longer using `chmod --reference`. In practice I don't think this will matter though.

benradf (approved):

aherrmann (approved):
Thank you @lamcw for contributing this!
And thanks @benradf for looking into this and the alternative approach!

LGTM

Also pinging @guibou who had the original use-case for preserving the executable bit. Please let us know if this change causes any issues for your use-case.

Issue #255, opened 2022-07-28
Add an `interpreter` target to python toolchain.


uri-canva:
`rules_python`'s `python_register_toolchains` has an `interpreter` target in its toolchains. Repository rules cannot access toolchains like regular rules can, and they use that target to ensure they're using the same interpreter as the toolchain.

See https://github.com/bazelbuild/rules_python/blob/9cdb4f3f6aded1ccc62a10d004f9927ccc72702f/README.md#toolchain-registration for example.

It would be good if `nixpkgs_python_configure` did the same thing here: https://github.com/tweag/rules_nixpkgs/blob/9f08fb2322050991dead17c8d10d453650cf92b7/toolchains/python/python.bzl#L80, for example by passing `build_file_content` to `nixpkgs_package`:

```
    build_file_content = """
package(default_visibility = ["//visibility:public"])
filegroup(
    name = "interpreter",
    srcs = ["bin/python"],
)
    """,
```

Repository rules of other language ecosystems have the same issue, but they don't seem to have a similar pattern (for example `rules_jvm_external` doesn't provide any way of specifying which java runtime to use other than the regular `PATH` / `JAVA_HOME`: https://github.com/bazelbuild/rules_jvm_external/blob/3957fdc382b5a404fccdde74b91c1e614e07e6bd/coursier.bzl#L206.

aherrmann:
Agreed, that sounds like a useful feature.

Worth noting, the current setup doesn't generate an executable file within Bazel's execroot, instead it points directly into the Nix store, see [here](https://github.com/tweag/rules_nixpkgs/pull/103) for the reasoning behind that. That said, creating a single `bin/python` symlink shouldn't be an issue.

Another thing worth noting, rules_nixpkgs provided toolchains are very careful to separate the external workspace that defines the `toolchain` from the external workspace(s) that hold the actual Nix imports. This is to ensure that the Nix dependencies are fetched lazily, and to support non-Nix use-cases (e.g. Windows). So, as proposed by @uri-canva, the `interpreter` target and the `bin/python` symlink probably have to sit within the workspace that defines the `py_runtime` target. That said, it should be possible for [`_nixpkgs_python_toolchain`](https://github.com/tweag/rules_nixpkgs/blob/9f08fb2322050991dead17c8d10d453650cf92b7/toolchains/python/python.bzl#L45) to generate a `defs.bzl` file that exports an `interpreter` like `rules_python` does.

aherrmann:
[...]

PR #261, opened 2022-08-16
Add an interpreter target to python toolchain

benradf:
My attempt at adding an `interpreter` target for the python toolchain as requested in #255. I think this maintains the separation between the toolchain workspace and the nix workspace you mentioned @aherrmann, but please let me know if not.


toolchains/python/python.bzl
```diff
@@ -36,20 +44,25 @@ toolchain(
...
+        python3_runtime = label_string(python3_runtime),
         exec_constraints = exec_constraints,
         target_constraints = target_constraints,
     ))
 
+    repository_ctx.file("defs.bzl", executable = False, content = """
+# Generated by rules_nixpkgs/toolchains/python/python.bzl
+interpreter = "{}"
+    """.format(repository_ctx.attr.python3_repo + ":bin/python"))
+
```
aherrmann, line 57: The case where `repository_ctx.attr.python3_repo` is empty doesn't seem to be handled here. In that case it should probably fall back to the python2 one.

aherrmann (changes requested):
Looks good, thank you!

One comment.

toolchains/python/python.bzl
benradf: Yeah, good point. Will do this now.

aherrmann (approved):
Thank you, looks good!

Closes #255

PR #263, opened 2022-08-24
Avoid restarting nixpkgs_package

lamcw:
TL;DR: pre-resolve all repository_ctx.path() before cp() calls.

See bazelbuild/bazel#4533 for background, repository rules in bazel will
be restarted if a depedency it requests is missing.

> When a Skyframe value is missing, fetching must be restarted, thus, in order to avoid doing duplicate work, it’s better to first request the Skyframe dependencies you need and only then start doing anything costly.
>
> https://github.com/bazelbuild/bazel/blob/0b485417a03b071038389609d0c66055db7ef636/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java#L150-L152

In the context of rules_nixpkgs, this can cause `nixpkgs_package` to be
restarted every time an additional label is referenced in the rule. This
is especially bad if the rule just happened to finish the slow `cp()`
calls and new labels are referred to after that.

As a workaround, we pre-resolve all labels at the top of repository_rule
and run the `cp`s towards to end.

This is a follow-up to #256

### DIFF
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -170,6 +170,10 @@ def _nixpkgs_package_impl(repository_ctx):
     for dir in nix_path:
         expr_args.extend(["-I", dir])
 
+    # Workaround to bazelbuild/bazel#4533 -- to prevent this rule being restarted after running cp,
+    # resolve all dependencies of this rule before running cp
+    #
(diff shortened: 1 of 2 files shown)


benradf:
Looks good! Thanks for submitting this further improvement @lamcw.

benradf (approved):

PR #265, opened 2022-09-14
Convert to flake

avdv:
* add compatibility layer for `nix-shell` users
* update nixpkgs to nixos-22.05
* use `flake.lock` instead of `nixpkgs.json`


tshaynik:
At the moment there are two separate `flake.nix` files, one in the project root and then a placeholder in `core` that would need to get synchronized to the same `rev` (presumeably these should always use the same `nixpkgs` commit?). So at the moment, to bump nixpkgs, you would have to run `nix flake update` in both the project root and in core to keep them in sync. Could it be useful to delete the `core/flake.nix` and make `core/flake.lock` a symlink? Then `nix flake update` in the project root would update core as well.

avdv:
Yes, good point! I'll see if that works...

tshaynik (approved):
This looks good to me! (I'm still very new to Tweag and to Bazel, so somebody else should probably look it over as well)
Things I like about your approach:
- using nix-compat to not breaking the traditional workflow
- ability to use a single invocation of `nix flake` tool to manage pinning nixpkgs
- actual use of this ruleset within Bazel does not depend on any experimental flake features

avdv:
> This looks good to me! (I'm still very new to Tweag and to Bazel, so somebody else should probably look it over as well)

Thank you for the review! Another pair of eyes would have been good, but I *accidentally* already pushed this onto master... If somebody has any objections to this, let's just fix it in another PR.

>  Things I like about your approach:
> 
>     * using nix-compat to not breaking the traditional workflow
> 
>     * ability to use a single invocation of `nix flake` tool to manage pinning nixpkgs
> 
>     * actual use of this ruleset within Bazel does not depend on any experimental flake features

:+1: we might switch to `nix flake` usage in the future, when flakes are no longer considered experimental.

aherrmann (commented):
I can only echo @tshaynik's comments, looks good, thank you!

PR #273, opened 2022-10-14
Fix java toolchain config to allow for use with Bazel 6

### LINKED ISSUE #272: `rules_nixpkgs` does not work with Bazel 6
**Describe the bug**
Bazel 6 is currently in pre-rerelease, but will see a stable release in Q4.
At the moment, running the tests with Bazel 6 breaks as a result to changes that have been made in the java toolchain.

**To Reproduce**
- start from this branch #270 to enable using unstable nixpkgs 
- Modify flake.nix to point to `nixpkgs-unstable`. Run `bazel test '//...'` and `cd core; bazel test '//...'`, and both should work fine
- Modify flake.nix devShell buildInputs from `bazel_5` to `bazel_6`.

``` shell
$ bazel test '//...
...'
ERROR: /home/tshaynik/.cache/bazel/_bazel_tshaynik/9b53471c1c450ee14443a850d69df9c9/external/nixpkgs_java_runtime_toolchain/BUILD.bazel:2:19: no such target '@remote_java_tools//:java_compiler_jar': target 'java_compiler_jar' not declared in package '' defined by /home/tshaynik/.cache/bazel/_bazel_tshaynik/9b53471c1c450ee14443a850d69df9c9/external/remote_java_tools/BUILD and referenced by '@nixpkgs_java_runtime_toolchain//:nixpkgs_java_toolchain_java11'
ERROR: /home/tshaynik/.cache/bazel/_bazel_tshaynik/9b53471c1c450ee14443a850d69df9c9/external/nixpkgs_java_runtime_toolchain/BUILD.bazel:2:19: no such target '@remote_java_tools//:jdk_compiler_jar': target 'jdk_compiler_jar' not declared in package '' defined by /home/tshaynik/.cache/bazel/_bazel_tshaynik/9b53471c1c450ee14443a850d69df9c9/external/remote_java_tools/BUILD and referenced by '@nixpkgs_java_runtime_toolchain//:nixpkgs_java_toolchain_java11'
``` 

**Expected behavior**
All tests should pass 

**Environment**
 - OS name + version: NixOS 22.05
 - Version of the code: tweag:rules_nixpkgs/nixVersions

**Additional context**
Here are some things I've found/tried so far:
- The java toolchain is currently a dependency of every part of the project, because of the doc generation, as referenced in #264, so the java toolchain not working breaks every component
- when running with Bazel 6, it errors out specifically complaining about `@remote_java_tools//:jdk_compiler_jar`, which is referenced in
`/toolchains/java/default_java_toolchain.bzl`
- `/toolchains/java/default_java_toolchain.bzl` and `/toolchains/java/default_java_toolchain.bzl` appear to be vendored from `rules_java` and modified to enable passing constraints, which we in order to enforce our constraint that the toolchain come from nixpkgs. This was brought in with #185, in https://github.com/tweag/rules_nixpkgs/pull/185/commits/618f20d1db84dfaefb5e5e8b41b2479b2ab9d169 .
- I'm not sure where that particular commit of Bazel or rules_java these two files are coming from, since they're copyright the Bazel authors 2017. The `default_java_toolchain.bzl` file has changed quite a bit in the past 5 years.
- I tried using files from newer versions of rules_java, and modifying the `default_java_toolchain` function and its callers to accept the `exec_compatible_with` and `target_compatible_with` constraints, but that wasn't working.

Now the naive thing that **actually worked** was to just go in to the existing `default_java_toolchain.bzl`and just delete all lines with references to `@remote_java_tools//:jdk_compiler_jar`. After that there's a similar error with `@remote_java_tools//:jdk_compiler_jar`. If I remove lines that reference that from `default_java_toolchain.bzl`, all tests pass!

**Questions on next steps**
- We could just remove those lines and have a quick fix. But it seems like there's a deeper issue about maintaining our modified, vendored files from `rules_java` in this repo, and that we should figure out!
- Could a PR  to `rules_java` adding the possibility to set those constraints be valuable? Presumably, with default options, this wouldn't be a breaking change to their API. If we did this, we wouldn't need to maintain modified versions of those 2 files in this repo anymore and could just use what's officially in `rules_java`, and we're already using rules_java as a dependency.

I'm down to keep working on this and get it finished up, but need some advice figuring out the best direction!

tshaynik:
Resolves #272 

This change allows for using rules_nixpkgs with Bazel 6, by fixing an issue caused by changes in how java toolchains are configured.

This works by pulling in versions of local_java_repository.bzl and default_java_toolchain.bzl from a newer version of Bazel, and reapplying the the changes necessary to enforce the constraint that nix be present.

TODO before merging:
- [x] Get #270 merged first
- [x] investigate MacOS failure on Bazel 6 from nixpkgs-unstable
- [x] discuss our strategy for managing the two vendored java toolchains files ([PR made to Bazel](https://github.com/bazelbuild/bazel/pull/16498) is orthogonal)
- [x] Revert to nixpkgs-22.05 and Bazel 5 before merging!!! Rebase and clean up commits

### DIFF
--- a/core/.bazelrc
+++ b/core/.bazelrc
@@ -12,6 +12,12 @@ build --java_language_version=11
 build --tool_java_runtime_version=nixpkgs_java_11
 build --tool_java_language_version=11
 
+# The following lines provide a workaround for a test failure in on the Java
+# toolchain on MacOS 11, documented here: https://github.com/tweag/rules_nixpkgs/issues/299.
+# They should be removed as soon as that issue has been resolved.
+build --enable_platform_specific_config
+build:macos --experimental_strict_java_deps=warn
+
 # User Configuration
 # ------------------
 try-import %workspace%/.bazelrc.local


tshaynik:
The java tests failure on MacOS seems to be resolved by setting `--strict_java_deps=warn`. I'm not sure if that's something that we want to add to the .bazelrc permanently, and still not sure what has changed that would have made this error start appearing.

tshaynik:
This is rebased, tested with Bazel 6, and now reset to Bazel 5, and ready for review.

I've created #299 to track the lingering Java test dependency issue on MacOS, and plan for the removal of the `--experimental_strict_java_deps=warn` workaround.

core/.bazelrc
```diff
@@ -11,6 +11,7 @@ build --java_runtime_version=nixpkgs_java_11
 build --java_language_version=11
 build --tool_java_runtime_version=nixpkgs_java_11
 build --tool_java_language_version=11
+build --experimental_strict_java_deps=warn
```
avdv, line 14: Could you add a short explanation here why this is necessary? And create a new issue which is linked in the comment please?

IIRC, this is only needed on macos, right? So, could we apply this only to this platform? (Maybe just leverage the [`--enable_platform_specific_config`](https://bazel.build/reference/command-line-reference#flag--enable_platform_specific_config) flag)

avdv (changes requested):

core/.bazelrc
tshaynik: Good call! I'll add this now.

avdv (approved):
Thanks for reminding me @tshaynik :smile: , looks good!

PR #276, opened 2022-11-07
Skip writing to the remote cache without API key

### LINKED ISSUE #243: Enable CI runs on external PRs
Enable CI runs on PRs coming from forks. As discussed in https://github.com/tweag/rules_nixpkgs/pull/236#issuecomment-1176396264 this is currently not possible because the BuildBuddy secret cannot be exposed to actions run from forks and the run fails due to missing credentials.

Ideally, we'd find a way to configure a read-only token for actions run from forks while using a read-write token for internal runs.

Short of that, we could simply run without BuildBuddy at all for actions triggered from forks.

avdv:
For foreign PRs (e.g. dependabot) access to the repository secrets is not permitted, so the API key for Buildbuddy ends up as empty.

This leads to an error later:
```
WARNING: BES was not properly closed
ERROR: The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=UNAUTHENTICATED, description=failed to parse API key: missing API Key, cause=null}' UNAUTHENTICATED: UNAUTHENTICATED: failed to parse API key: missing API Key UNAUTHENTICATED: UNAUTHENTICATED: failed to parse API key: missing API Key
```

This PR is a simpler alternative to the approach taken in #248: instead of artificially drawing a line between pull_request and push events this change will just not try to write to the remote cache if the API key is not set (for whatever reason). 

Update: I just noticed that dependabot actually does create a branch in this repository, which triggers a push. But still, it has no access to the repository secrets. So the approach in #248 would not fix this, IMO.

Fixes #243


bryanhonof (approved):
Looks good to me. Could we perhaps also make this a successor to #248? I believe I've overengineered that one. 😅

PR #277, opened 2022-11-07
Skip writing to the remote cache with no API key - no. 2

avdv:
In PR #276 I failed to adapt the second "Configure" step of the "test-examples" job. Do this now.


tshaynik (approved):
LGTM!

PR #279, opened 2022-11-14
Replace Buildkite CI badge with Github badge

avdv:
Currently, the badge says CI is failing and links to builds on Buildkite from March 2021.

Should we add a badge for upstream Buildkite builds maybe?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Nixpkgs rules for Bazel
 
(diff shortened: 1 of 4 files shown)


aherrmann (approved):
Thank you!

avdv:
> You mean bazel CI? Do we have a rules nixpkgs pipeline there? I thought it was only rules haskell.

Oh, my bad. I thought there was a `.bazelci` folder, but it was the `.bazelrc.ci` file.

PR #286, opened 2022-11-30
examples: mkShell -> mkShellNoCC

evertedsphere:
Given that using platform-provided CC toolchains is one of the major sources of inhermeticity in Bazel usage, this would serve to underscore what `rules_nixpkgs` brings to a build setup.

Everything should still work.


aherrmann:
Looks like the cc_with_deps example [fails with a glibc version issue](https://github.com/tweag/rules_nixpkgs/actions/runs/3583870861/jobs/6029819460#step:5:274).
```
/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33'
```

evertedsphere:
@aherrmann it turns out the `rules_nixpkgs`-provided CC toolchain wasn't being used at all. Adding a `--crosstool-top` setting to `.bazelrc` fixes it.

AlexeiDrake (approved):

Issue #289, opened 2022-12-01
Document use of CC toolchain with --incompatible_enable_cc_toolchain_resolution


aherrmann:
**Is your feature request related to a problem? Please describe.**
The current documentation for the Nix provided CC toolchain only mentions the `--crosstool_top` flag. However, Bazel is moving the CC toolchain forward to integrate with platforms & toolchains. This feature can be enabled with the [`--incompatible_enable_cc_toolchain_resolution`](https://github.com/bazelbuild/bazel/issues/7260) flag. The Nix provided CC toolchain is compatible with the platforms&toolchains mechanism.

**Describe the solution you'd like**
The CC rules in Bazel are still not fully migrated, meaning that `--crosstool_top` may still be necessary in some use-cases. However, users that want to use platforms & toolchains should be made aware that the Nix provided CC toolchain is compatible with this and that they can enable this mode with the above mentioned flag.

**Additional context**
We have found that even with the use of `--incompatible_enable_cc_toolchain_resolution` and the correct platform configuration there seem to be some hard-coded references to `local_config_cc` in Bazel. This means that when a project configures a Nix provided CC toolchain under a different name, e.g. `nixpkgs_config_cc`, then the inhermetic default toolchain `local_config_cc` may still be fetched in used some contexts.

PR #294, opened 2022-12-20
doc: add a guide covering basic rules_nixpkgs use + Nix prerequisites

### LINKED ISSUE #269: Create user-guide documentation
Thanks to the recent work on Stardoc, rules_nixpkgs now has complete generated API documentation in its `README.md`. However, API documentation only serves as a reference and doesn't necessarily help people who are just getting started.

rules_nixpkgs should have user-guide documentation that explains when rules_nixpkgs is useful, its purpose and motivation, and then guides a new user through the process of setting it up and using it.
We already have a couple of examples in the repository that also serve as integration tests and have the advantage that their correctness is tested on CI. We could incorporate these examples into a user-guide.

evertedsphere:
Add an initial version of a rules_nixpkgs guide (fix #269)

The target audience is Bazel users new to Nix, so I've taken care to not assume Nix-related knowledge and to introduce the minimum amount of information required to be able to start using the ruleset.

For the purposes of this version, we only target Linux; #371 tracks macOS support.


guide.md
```diff
@@ -0,0 +1,324 @@
+---
+title: "Fully hermetic Bazel: transitioning to rules_nixpkgs"
+---
+
+Bazel is a modern build system valued for performant, correct, and hermetic
+builds. We have a sizeable number of Bazel users here at Tweag, and it's
+something we are passionate about and often advocate for the use of on the
+complex polyglot projects we help clients with.
+
+It's not without its problems, though, and the one we're interested in today is
```
aherrmann, line 10: > and the one we're interested in today is

Good for the blog post, but I'd leave it out for the user-guide.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+from Bazel, allowing it to provision external dependencies via the Nix package
+manager. Nix provides strong hermeticity and reproducibility guarantees for all
+build products, which are defined in a language also called Nix. A sizeable
+collection of such package definitions is available in the Nixpkgs repository,
+which can then be used freely from Bazel.
+
+# A basic Bazel project
+
+Let's start with a simple example, on the Bazel "happy path": building a simple
+C/C++ project. The project structure we'll work with is available [here] if
```
aherrmann, line 33: > is available [here]

Link missing.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+executables in the PATH and sets up a C/C++ compiler toolchain called
+`@local_config_cc` that "should work". 
+
+This is problematic, and not just because of academic concerns regarding the
+precise meaning of "hermeticity": build products output by different versions
+of the same compiler can vary greatly, let alone between different compilers.
+Bazel, by default, chooses GCC on a Linux machine and Clang/LLVM on a macOS
+machine, which can mean that the same bit of hot code, say, vectorizes
+correctly on one platform but not on the other due to some difference between
+the two compilers, resulting in wildly different performance characteristics.
```
aherrmann, line 95: Uncontrolled system library versions, e.g. glibc version, could also be a compelling issue.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+
+There are Bazel-native ways to mitigate these issues, but they fall short of true
+hermeticity: for instance, `--incompatible_strict_action_env` runs actions in
+an environment that does not inherit a `PATH` from the shell environment, but
+instead [sets it to][path-link] the static value `/bin:/usr/bin:/usr/local/bin`.
+This means that as long as your compiler tools are in one of those paths, which
+they almost certainly are on \*nix systems, nothing changes.
```
aherrmann, line 102: Generally true, but for the builtin C/C++ toolchain example this is irrelevant. The default C/C++ toolchain setup happens in a repository rule which is unaffected by `--incompatible_strict_action_env`.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+environment ourselves so that the implicit dependencies end up being the same
+every time. To do so, we run all builds in sandboxes or VMs that all share the
+exact same configuration, for instance, in identical Docker containers.
+
+The other solution is to eliminate the dependency on the global environment by
+defining your own compiler toolchain, providing paths to executable binaries that you
+provision and version yourself. Bazel can then be instructed to use that
+instead of the default `@local_config_cc`. This can be done by checking all the
+relevant executables (`cc`, `ld`, `as`, ...) into your repository yourself, as
+is done within Google for instance.
```
aherrmann, line 117:
```suggestion
relevant executables (`cc`, `ld`, `as`, ...) into your repository yourself. 
```
Not sure if they've documented that properly and publically anywhere. I've been told that's the case, but that's just hearsay.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+In both cases, the build depends on non-code elements (Docker containers,
+executable binaries) for which versioning and maintenance is more complex than
+for ordinary code. Extra steps must now be followed when these environments
+have to be changed or when new releases are cut. Either these configuration
+files have to be executed at runtime, necessitating long setup processes for
+each executor, or one caches the build products. In the latter case, large
+binary blobs have to be managed and moved from machine to machine, often in the
+hundreds of gigabytes for Docker images.
+
+The r_n approach is a variation on the second, where instead of managing system
```
aherrmann, line 128:
```suggestion
The rules_nixpkgs approach is a variation on the second, where instead of managing system
```
I'd avoid such abbreviations.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+exact same configuration, for instance, in identical Docker containers.
+
+The other solution is to eliminate the dependency on the global environment by
+defining your own compiler toolchain, providing paths to executable binaries that you
+provision and version yourself. Bazel can then be instructed to use that
+instead of the default `@local_config_cc`. This can be done by checking all the
+relevant executables (`cc`, `ld`, `as`, ...) into your repository yourself, as
+is done within Google for instance. 
+
+In both cases, the build depends on non-code elements (Docker containers,
```
aherrmann, line 119: The Docker case has another issue.
1. It's coarse grained - we don't know exactly which elements of the Docker image we depend on, so, if we change it, we have to invalidate the entire build.
2. Most Dockerfiles out there are not reproducible, e.g. they invoke `apt-get update`. So, if we need to rebuild the Docker image because we want to add just one library, we may incidentally also change other dependencies.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+executable binaries) for which versioning and maintenance is more complex than
```
aherrmann, line 120: The binary blobs approach also has the problem that it increases the trusted base. I.e. we need to trust that the checked in binary is really the version we want and has the behavior we want.

guide.md
aherrmann, line 128: I'd characterize it as a different approach rather than a variation. The reason is that it enables fine grained dependency tracking on system dependencies and toolchains. E.g. if we only add a system library dependency imported through rules_nixpkgs, it won't affect the CC toolchain and we won't have to invalidate previously built CC targets.

One reason I'd make that distinction is that a variation of the Docker approach is indeed possible with Nix. We could use Nix Docker tools to build a Docker image with the build environment in a reproducible way. However, this still has the issue of being coarse grained, since Bazel doesn't track precisely which parts of the image the build depends on.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+nixpkgs_git_repository(
+    name = "nixpkgs",
+    revision = "22.05",
+    sha256 = "",
+)
+```
+
+# A hermetic CC toolchain
+
+Now we can now define a CC toolchain that will be provisioned by Nix using
```
aherrmann, line 182:
```suggestion
Now we can define a CC toolchain that will be provisioned by Nix using
```

guide.md
```diff
@@ -0,0 +1,324 @@
...
+
+This will select the default `stdenv.cc` package as the compiler, drawing
+auxiliary tools from `stdenv.cc.bintools`. However, one can write a little bit of
+Nix to configure this:
+
+```bazel
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_cc_configure")
+nixpkgs_cc_configure(
+  repository = "@nixpkgs",
+  nix_file_content = "(import <nixpkgs> {}).gcc11",
```
aherrmann, line 201: Wouldn't `attribute_path = "gcc11"` achieve the same?

guide.md
```diff
@@ -0,0 +1,324 @@
...
+architecture, and have everything work seamlessly.
+
+With this, we now have a fully hermetic CC toolchain that will behave correctly
+across platforms. It remains to tell Bazel to use it by default, instead of
+`@local_config_cc`. To do so, we put the following into `.bazelrc` at the root
+of the repository:
+
+```
+build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
+run   --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
```
aherrmann, line 219:
```suggestion
```
`run` inherits from `build`.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+build --crosstool_top=@nixpkgs_config_cc//:toolchain
```
aherrmann, line 220: Maybe worth mentioning [`https://github.com/bazelbuild/bazel/issues/7260`](https://github.com/bazelbuild/bazel/issues/7260) as well?

guide.md
```diff
@@ -0,0 +1,324 @@
...
+
+# Making Nix optional
+
+It is entirely possible to make the use of Nix optional in a Bazel setup. To do
+so, we use `bazelrc` configs to only enable Nix-specific configuration when a
+`--config` flag is provided, like so:
+
+```
+build:nix --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
+run:nix   --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
```
aherrmann, line 244:
```suggestion

```

guide.md
```diff
@@ -0,0 +1,324 @@
...
+nix-shell> exit
+> hello
+hello: command not found
+```
+
+There's still a few loose ends in our setup that Nix developer environments can
+help tie up. First, the version of Bazel we've been using so far, has been
+unspecified. The general practice is to simply use one's distribution's package
+manager, something that doesn't deliver consistent Bazel versions across major
+OS platforms. Second, we haven't _disabled_ Bazel's deep-rooted tendency to
```
aherrmann, line 279: I think the more common practice outside of Nix is to use bazelisk, which can be configured to use a consistent Bazel version for the project.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+> hello
+hello: command not found
+```
+
+There's still a few loose ends in our setup that Nix developer environments can
+help tie up. First, the version of Bazel we've been using so far, has been
+unspecified. The general practice is to simply use one's distribution's package
+manager, something that doesn't deliver consistent Bazel versions across major
+OS platforms. Second, we haven't _disabled_ Bazel's deep-rooted tendency to 
+depend on the global environment. Indeed, you can rename `.bazelrc` and rerun
```
aherrmann, line 280: Does the example above actually work with a vanilla Bazel bindist? Using rules_nixpkgs with vanilla Bazel (i.e. non-Nix- provided Bazel) is not a supported use-case.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+      let pkgs = import nixpkgs { inherit system; };
+      in
+      {
+        devShells.default = with pkgs; mkShellNoCC {
+          name = "rules_nixpkgs_shell";
+          packages = [ bazel_5 bazel-buildtools cacert gcc nix git ];
+        };
+      });
+}
+```
```
aherrmann, line 308: I fear that much boilerplate will require some explanation. Maybe you could reference one of the flake templates and then explain where to add the bazel stuff?

guide.md
```diff
@@ -0,0 +1,324 @@
...
+    flake-utils.url = github:numtide/flake-utils;
+  };
+  outputs = { nixpkgs, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let pkgs = import nixpkgs { inherit system; };
+      in
+      {
+        devShells.default = with pkgs; mkShellNoCC {
+          name = "rules_nixpkgs_shell";
+          packages = [ bazel_5 bazel-buildtools cacert gcc nix git ];
```
aherrmann, line 304: Why `gcc`?

guide.md
```diff
@@ -0,0 +1,324 @@
...
+        };
+      });
+}
+```
+
+Now developers on different platforms can enter a development environment by
```
aherrmann, line 310: Maybe mention `direnv` integration and how to set up `.envrc`. That makes it much more convenient to use this.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+running `nix develop`. We've also removed the "default" system-wide CC
+toolchain (using `mkShellNoCC` instead of the more common `mkShell`), so
```
aherrmann, line 312: > We've also removed the "default" system-wide CC
> toolchain (using `mkShellNoCC` instead of the more common `mkShell`),

I think Nix specifics like these will require a bit more expanding.

guide.md
```diff
@@ -0,0 +1,324 @@
...
+`@local_config_cc` will no longer work:
+
+```
+> bazel build //src:hello-world 
+INFO: Build options --crosstool_top, --host_crosstool_top, --host_platform, and 1 more have changed, discarding analysis cache.
+[snip]
+Auto-Configuration Error: Cannot find gcc or CC; either correct your path or set the CC environment variable
+```
```
aherrmann, line 320: The guide should make it very clear when a snippet or described approach is considered "correct" and when it is "wrong". I'd also err on the side of "correct" first and "wrong" after (if at all). The reasoning here is this: Guides are often not read top-to-bottom carefully, but rather jumped around in or taken as reference. So, it should be very clear whether a given snippet or approach is safe to copy or not. E.g. the Rust book does this nicely with big visual indications.

aherrmann (commented):
Thank you! I did a first pass and added some comments.

guide.md
evertedsphere: Oops, that wasn't supposed to be left in.

guide.md
layus: I also think you should only provide working snippets. Also, are you sure that running in mkShellNoCC will not default to system gcc on non NixOS configurations ?
At this point I am a bit lost in what is the current config, and why. This should probably move to a separate section, and keep the guide focussed on the intended, canonical way to use it with the benefits.

guide.md
evertedsphere: Added.

guide.md
evertedsphere: Added.

guide.md
evertedsphere: Added.

guide.md
evertedsphere: Good point, I've clarified this.

guide.md
evertedsphere: Right, I see what you both mean. I'll rework this section with that feedback in mind.

guide.md
evertedsphere: I'll take it out since it's not supported (and getting a Nix-provided Bazel is trivial once you have Nix, so it shouldn't prevent anyone from following this guide.)

guide.md
evertedsphere: Done.

guide.md
evertedsphere: Done.

guide.md
evertedsphere: I took out the error, and explained that bit in detail in words instead.

guide.md
evertedsphere: Took this out.

guide.md
evertedsphere: Added.

guide.md
evertedsphere: (Section taken out, see below.)

guide.md
evertedsphere: I've also added instructions for that (i.e. a basic `nix-shell -p bazel_5`) which I think should be enough.

guide.md
evertedsphere: I can't seem to get `--incompatible_enable_cc_toolchain_resolution` to work on this repo, even if I specify the platform name manually on the command line, presumably because I haven't told it to use the `@nixpkgs_config_cc` toolchain. Renaming it to `@local_config_cc` works (and indeed simplifies some of the guide). 

However, if I make that change, I don't know how to adapt the "Making Nix optional" section (or if it needs to be adapted at all: will toolchain resolution silently fall back to the default toolchain?)

guide.md
evertedsphere: I couldn't find a flake template that used flake-utils (which we need for `eachDefaultSystem` in order to have the cross-platform shell that we're advertising) and was sufficiently similar to this that one could just add in the Bazel stuff without complications.

I've cleaned it up a bit and added some comments inline.

guide.md
```diff
@@ -0,0 +1,350 @@
...
+ecosystems where there is widespread adoption of a package manager, this isn't
+much of an issue --- for instance, `pip` dependencies are taken care of by
+`rules_python`. However, in the case of C/C++ projects or with library headers,
+third-party dependencies can still be difficult to manage.
+
+The rules_nixpkgs ruleset enables the use of the extensive Nixpkgs package set
+from Bazel, allowing it to provision external dependencies via the Nix package
+manager. Nix provides strong hermeticity and reproducibility guarantees for all
+build products, which are defined in a language also called Nix. A sizeable
+collection of such package definitions is available in the Nixpkgs repository,
```
mrkkrp, line 25: Given the general introductory tone of this document, it feels like mentions of technologies and repos like Nixpkgs should have links?

guide.md
```diff
@@ -0,0 +1,350 @@
...
+tag for a stable release, which ensures that there has been especially
+stringent verification that packages in it will function correctly individually
+and play well together. Here we use the 22.05 release:
+
+```bazel
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_git_repository")
+nixpkgs_git_repository(
+    name = "nixpkgs",
+    revision = "22.05",
+    sha256 = "",
```
mrkkrp, line 206: Should there be anything about specifying `sha256`, how to find it and why it might be a good idea?

guide.md
```diff
@@ -0,0 +1,350 @@
...
+
+```
+build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
+build --crosstool_top=@nixpkgs_config_cc//:toolchain
+```
+
+Now `bazel build` and `bazel run` should use this toolchain by default for
+C/C++ compilation:
+```
+> bazel build //src:hello-world
```
mrkkrp, line 256: This snippet does not "prove" that a different toolchain is indeed used. Perhaps building with `--toolchain_resolution_debug` could illustrate the switch better?

guide.md
mrkkrp: So far the explanations were very detailed, but here the user is left pondering what these two options really mean. Maybe add a few sentences about them, even if it is in a hand-wavy way?

guide.md
```diff
@@ -0,0 +1,350 @@
...
+```
+> bazel run //src:hello-world --config=nix
+...
+```
+
+# Working from a Nix-provided developer environment
+
+The `nix-shell` command we used to get a Nix-provided Bazel was the simplest
+example of a Nix developer environment, but they can provide developers on a
+complex project with a lot more convenience. We can create development
```
mrkkrp, line 289:
```suggestion
example of a Nix developer environment, but it can be more sophisticated.
We can create development
```

guide.md
```diff
@@ -0,0 +1,350 @@
...
+environments that all contributors on a project can use with zero setup. These
+shells provide consistent versions of libraries for whichever languages are
+used in the project as well as for system dependencies.
+
+We haven't _disabled_ Bazel's deep-rooted tendency to depend on the global
```
mrkkrp, line 294: This is not entirely true, or at least confusing for me. One could argue that we partially disabled it earlier when we used

```
build --host_platform=@io_tweag_rules_nixpkgs//nixpkgs/platforms:host
build --crosstool_top=@nixpkgs_config_cc//:toolchain
```

Maybe rephrase it to make clear that without special care and options like the ones above Bazel might be influenced by its environment?

guide.md
```diff
@@ -0,0 +1,350 @@
...
+`mkShell` function creates a development shell with a given set of available
+dependencies _and_ a default CC toolchain, so we've replaced it with
+`mkShellNoCC`, which, as the name suggests, does the same job but without
+implicitly making a CC toolchain available. This means it's no longer possible
+to unintentionally use `@local_config_cc`, as that toolchain won't be able to
+find the global binaries it wants and will fail to work.
+
+Now developers on different platforms can enter a development environment by
+running `nix develop`, and this environment will be identical across platforms.
+This can be made even more seamless by using [direnv](direnv), which has support for Nix flakes:
```
mrkkrp, line 341: This is the first time you mention flakes explicitly, and it is confusing for someone who is not familiar with Nix, i.e. your target audience IIUC.

guide.md
```diff
@@ -0,0 +1,350 @@
...
+```
+$ echo `use flake` > .envrc
+$ direnv allow
+```
+With this, this environment will be entered automatically when in this directory.
```
mrkkrp, line 346: Too many "this"?

guide.md
```diff
@@ -0,0 +1,350 @@
...
+environment. Indeed, you can rename `.bazelrc` and rerun `bazel run`, and
+everything will work just fine: Bazel will simply fall back to
+`@local_config_cc`, because a C++ compiler toolchain is still globally
+available.
+
+We can provide a Nix developer environment that hermetically provides the specific
+dependencies that developers need and nothing else:
+
+```nix
+{
```
mrkkrp, line 304: This snippet is incomplete at least in the sense that you do not say how to name a file like this and where to put it, which is important. Another thought might be that a `shell.nix` might be a more straightforward entry point than `flake.nix`.

mrkkrp (commented):
I left some comments. Great job, I enjoyed going through this! I feel like the beginning of the guide is more polished and you probably spent more time working on it and its wording. There is a noticeable difference in sophistication of writing, which gets simpler and more sketchy towards the end. It would be nice to bring the second half to the same level as the first.

aherrmann (commented):
Thank you, it's a nice read, good work!
I've added a couple comments.

guide.md
evertedsphere: Good catch, thanks.

Re: a `shell.nix`, I also considered that approach, but the argument against that is that 1. shell.nix relies on a global `<nixpkgs>` by default and 2. even with some kind of nixpkgs pinning, updating that nixpkgs version is more involved than the `nix flake update` interface.

guide.md
evertedsphere: Good idea, added.

guide.md
evertedsphere: Added a note

guide.md
evertedsphere: Done.

guide.md
evertedsphere: I reworded this to skip "flakes" and just mention the flake.nix file.

evertedsphere: I also added a link to our "intro to flakes" blog posts.

guide.md
```diff
@@ -0,0 +1,350 @@
+# A guide to rules_nixpkgs
+
+TODO: does use flake work with vanilla direnv? nix-direnv is just an improvement over that, i believe
```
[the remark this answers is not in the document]
evertedsphere: Thanks, I've added the `source_url` directive and added a note about the versions. Since nix-direnv significantly improves upon the support present in vanilla direnv in ways users are likely to notice very soon (caching and automatic GC roots), it makes sense to start users off with it in my opinion.

guide.md
```diff
@@ -0,0 +1,350 @@
...
+```
+
+# The lack of hermeticity
+
+Bazel is prized for its hermetic and reproducible builds, but this example
+already shows that that isn't completely accurate: we haven't told Bazel what
+C++ compiler to use, or where to find it and all the auxiliary tools it relies
+on (linker, assembler, and the like). It simply looks for appropriately-named
+executables in the PATH and sets up a C/C++ compiler toolchain called
+`@local_config_cc` that "should work".
```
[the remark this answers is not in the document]
evertedsphere: Good point, fixed.

guide.md
```diff
@@ -0,0 +1,350 @@
...
+greatly, let alone between different compilers. Bazel, by default, chooses GCC
+on a Linux machine and Clang/LLVM on a macOS machine, which can mean that the
+same bit of hot code, say, vectorizes correctly on one platform but not on the
+other due to some difference between the two compilers, resulting in wildly
+different performance characteristics. System library versions are not
+controlled, so build products will depend on whatever versions of, e.g. `glibc`
+are available on the host.
+
+There are Bazel-native ways to mitigate these issues in certain cases, but they
+fall short of true hermeticity.
```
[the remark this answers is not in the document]
evertedsphere: I wrote that as a replacement for the previous remark about `--incompatible_strict_action_env`, which you noted in your previous review didn't affect CC compilation. I'll update this keeping what you said in mind.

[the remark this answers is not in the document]
evertedsphere: I removed that bit and folded it into the next paragraph.

guide.md
```diff
@@ -0,0 +1,350 @@
+# A guide to rules_nixpkgs
+
+TODO: does use flake work with vanilla direnv? nix-direnv is just an improvement over that, i believe
+TODO: use the same nixpkgs.nix for nixpkgs_git_repository and shell.nix
```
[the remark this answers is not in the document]
evertedsphere: I've implemented that here: https://github.com/tweag/rules_nixpkgs_cc_template/blob/main/WORKSPACE#L22

guide.md
```diff
@@ -0,0 +1,350 @@
...
+TODO: is the flake shell pure? because if this is run on a non-nixos system, and /usr/bin/gcc continues to be visible, that messes things up
```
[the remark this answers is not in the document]
evertedsphere: If the Nix shell isn't pure, some of the stuff I've written about `@local_config_cc` not working when in the shell and having that extra insulation from accidentally using the non-hermetic toolchain becomes false, doesn't it?

guide.md
```diff
@@ -0,0 +1,350 @@
...
+INFO: Elapsed time: 0.377s, Critical Path: 0.22s
+INFO: 6 processes: 4 internal, 2 linux-sandbox.
+INFO: Build completed successfully, 6 total actions
+```
+
+# Making Nix optional
+
+It is entirely possible to make the use of Nix optional in a Bazel setup. To do
+so, we use `bazelrc` configs to only enable Nix-specific configuration when a
+`--config` flag is provided, like so:
```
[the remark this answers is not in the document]
evertedsphere: Hadn't thought about that, I'll make a note of this. 

As for defining a non-Nix toolchain, the `rules_cc_toolchains()` call takes care of that, right?

guide.md
evertedsphere: Added some links, but I'm not sure what the best place to link for Nix-the-package-manager/Nix-the-language/etc is... maybe nix.dev?

guide.md
```diff
@@ -0,0 +1,350 @@
...
+
+Now developers on different platforms can enter a development environment by
+running `nix develop`, and this environment will be identical across platforms.
+This can be made even more seamless by using [direnv](direnv), which has support for Nix flakes:
+```
+$ echo `use flake` > .envrc
+$ direnv allow
+```
+With this, this environment will be entered automatically when in this directory.
+
```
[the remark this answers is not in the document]
evertedsphere: Made a first pass at this.

aherrmann:
IIUC https://github.com/tweag/rules_nixpkgs_cc_template is the worked example for this user-guide, correct? Could you pull it into the rules_nixpkgs repository and add a test for it to CI? (Probably best in its own PR). rules_haskell does the same with [its tutorial](https://github.com/tweag/rules_haskell/tree/ae0e85988e5cb6080bcad5ae262c98c8fcd78896/tutorial).

That makes sure that the example is tested in CI, and that changes to the example can happen along-side the relevant changes to the guide and vice-versa.

guide.md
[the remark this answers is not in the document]
aherrmann: I added some comments there
https://github.com/tweag/rules_nixpkgs_cc_template/commit/a3d4e4011f610b4207c1353f4f2c03bb94c31383#r99606471
https://github.com/tweag/rules_nixpkgs_cc_template/commit/b5837f5a962d47357b6c46ab2987e1f201de025c#r99606695

guide.md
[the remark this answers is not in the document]
aherrmann: True, best to double check with `#nix`. I don't see `--pure` flag similar to the `nix-shell` one [on `nix develop`](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-develop.html), only `--impure`, but IIUC that is about Nix evaluation rather than the resulting shell environment.

guide.md
```diff
@@ -0,0 +1,440 @@
...
+define a development shell is with the commonly-used `mkShell` Nix function,
+which creates a development shell with a given set of available dependencies
+_and_ a default CC toolchain. We've replaced it with `mkShellNoCC`, which, as
+the name suggests, does the same job but without implicitly making a CC
+toolchain available. This means it's no longer possible to unintentionally use
+`@local_config_cc`, as that toolchain won't be able to find the global binaries
+it wants and will fail to work.
+
+Now developers on different platforms can enter a development environment by
+running `nix develop`, and this environment will be identical across platforms.
```
aherrmann, line 381: > this environment will be identical across platforms

Nit: I think I know what you're getting at, but it's a bit ambiguous. Is the user-guide only targeting Linux or is it also expected to work on MacOS? The reason I'm asking: Arguably "across platforms" would also mean across Linux and MacOS. But, in that case the environment is not identical, on MacOS it'll include Darwin executables instead of Linux ones.

guide.md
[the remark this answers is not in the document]
aherrmann: Thank you, that looks great!

guide.md
```diff
@@ -0,0 +1,440 @@
+# A guide to rules_nixpkgs
+
+TODO: is the flake shell pure? because if this is run on a non-nixos system, and /usr/bin/gcc continues to be visible, that messes things up
+TODO: document nixpkgs.nix
+
+---
+
+Bazel is a modern build system valued for performant, correct, and hermetic
+builds. It has a few shortcomings, one of which is that that Bazel is only
+hermetic if certain conditions are met, and they aren't in many people's build
```
aherrmann, line 10:
```suggestion
Bazel is a modern build system valued for fast and correct builds.
Fast, thanks to advanced caching and distributed builds.
Correct, in that build artifacts always reflect the state of their declared inputs. In other words, you should never have to run `bazel clean`.
Bazel's caching and distributed build features depend on correct build definitions.

Bazel build actions run in isolation to ensure that their build definitions are correct and capture all requirements.
This is referred to as "hermeticity": A build is hermetic if there are no external influences on it. However, Bazel is only
hermetic if certain conditions are met, and they aren't in many people's build
```
This uses some Bazel jargon, like "hermetic", without explanation. Maybe it could be expanded a bit to be more approachable to people not embedded in that jargon, yet.

guide.md
```diff
@@ -0,0 +1,440 @@
...
+setups. In practice, Bazel build products fail to be reproducible because Bazel
+silently depends on various things present in the global environment, like a
+C++ compiler. This ties into another issue, which is that external dependencies
```
aherrmann, line 13: Perhaps the external deps thing should be its own paragraph. Yes, it's related, and rules_nixpkgs covers both, but it's still a somewhat separate concern.

guide.md
evertedsphere: I've changed this to "consistent" instead of "identical". 

As for the first question: I wrote the guide with Linux in mind, but is there a reason why the guide wouldn't work on macOS?

evertedsphere: I've changed this to "consistent" instead of "identical". 

As for the first question: I wrote the guide with Linux in mind. Not sure if it would work elsewhere, but perhaps we can investigate macOS compatibility and try to make it work if we have a reasonable number of macOS users.

guide.md
evertedsphere: Done

guide.md
evertedsphere: Good point, done (and similarly elsewhere throughout the document)

evertedsphere:
Reordered the whole guide/rewrote large parts of it to accommodate the `nixpkgs.nix` flow with `nixpkgs_local_repository` by introducing `nix develop` and `direnv` first.

evertedsphere:
(Still have to add the note about non-Nix toolchains.)

guide.md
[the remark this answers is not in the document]
evertedsphere: I've added a note about this; could you check that it makes sense? (I wasn't really sure what the "default bindist" specifically referred to.)

guide.md
aherrmann: Makes sense. Focusing on Linux first seems perfectly fair and I'm happy to leave MacOS for follow-up work. In that case it should just be stated somewhere that this guide is only targeting Linux and we'd also need an issue on rules_nixpkgs to track the MacOS support.

guide.md
```diff
@@ -0,0 +1,559 @@
...
+shell, this will take care of all of them in one go, and update the `flake.lock`
+to include all of the changes.
+
+## Using `direnv` with Nix
+
+The use of this feature can be made even more seamless by using
+[direnv](direnv), which has support for the `flake.nix` file we used to describe
+the development shell.
+
+Put the following in a file called `.envrc` at the root of the repo:
```
aherrmann, line 335: The `.envrc` file is currently missing in `examples/cc-template`. Was the intention to include it?

guide.md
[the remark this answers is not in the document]
aherrmann: Thanks, yes, that makes sense. The exact shape of the "default bindist" depends on the language. With cc its implicit since it's builtin to Bazel. With other languages it would be your `rules_XXX_toolchains` call, or similar that sets this up, e.g. [here for Go](https://github.com/tweag/rules_nixpkgs/blob/56b9cff9d175b916abdb36920a669b1face49e04/examples/toolchains/go/WORKSPACE#L45).

examples/cc-template/flake.lock
```diff
@@ -0,0 +1,25 @@
+{
```
aherrmann, line 1: This needs an update. I'm getting changes to this file when I run `nix develop`.

examples/cc-template/WORKSPACE
```diff
@@ -0,0 +1,36 @@
+# load the http_archive rule itself
```
aherrmann, line 1: Once you rebase on master you can find an easy example of how to test this on CI [here](https://github.com/tweag/rules_nixpkgs/blob/691582fd6712502b38e7ef6a488203edc79143b8/.github/workflows/workflow.yaml#L122-L126).

aherrmann (commented):
That looks very good, thank you! A couple small comments, but nothing major.
The code looks good and I like the restructured form. Good work!

guide.md
evertedsphere: Yep, thanks for the catch. (`.envrc` is in my global gitignore.)

guide.md
```diff
@@ -0,0 +1,559 @@
...
+square brackets, as you might expect, set off a _list_ in Nix, with elements
+separated (admittedly a bit oddly) by whitespace. If you need packages in your
+development shell that aren't listed there already (compilers, development
+tools, editors, anything at all) you can add them to the `packages` list, and
+they'll become available within the shell when you enter it. In order to search
+for the package you want in Nixpkgs, there is a [package
+search](https://search.nixos.org/packages) function available on the Nixpkgs
+website.
+
+(TODO: this is false without --ignore-environment, which we haven't put into the
```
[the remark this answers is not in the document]
evertedsphere: Not sure what to do here; `--ignore-environment` seems to not reset everything, and even if it did, it can't be used (directly) from direnv. I've asked again on #nix on Slack.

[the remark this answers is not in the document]
evertedsphere: The only idea I have at this point is using `flake-compat` and getting a shell.nix and doing `use nix --pure`, which _is_ supported.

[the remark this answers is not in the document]
evertedsphere: I've rewritten the section, but it feels a little bit like lying by omission if we talk a big game about how good Nix shells are for testing software interactively without unnecessary dependencies and then provide a shell setup that isn't.

The hermeticity/correctness of the rules_nixpkgs-based Bazel build itself doesn't depend on this, though, as far as I know, so I suppose this setup won't actually create any incorrect builds. We can leave it at this for a first draft until we find a better solution.

examples/cc-template/WORKSPACE
```diff
@@ -0,0 +1,36 @@
...
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# load rules_nixpkgs
+http_archive(
+    name = "io_tweag_rules_nixpkgs",
+    strip_prefix = "rules_nixpkgs-9f08fb2322050991dead17c8d10d453650cf92b7",
+    urls = ["https://github.com/tweag/rules_nixpkgs/archive/9f08fb2322050991dead17c8d10d453650cf92b7.tar.gz"],
+    sha256 = "46aa0ca80b77848492aa1564e9201de9ed79588ca1284f8a4f76deb7a0eeccb9",
+)
```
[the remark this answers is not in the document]
evertedsphere: This does point to the commit at the tip of 0.9.0; for whatever reason, when I switch this to
```
http_archive(
    name = "io_tweag_rules_nixpkgs",
    strip_prefix = "rules_nixpkgs-0.9.0",
    urls = ["https://github.com/tweag/rules_nixpkgs/archive/refs/tags/v0.9.0.zip"],
    sha256 = "a8f4cc94757f19a0a1ce95245f7a87f294cfbce9859ed55ffaeeacf6f0d0af16",
)
```
I get errors about illegal symlinks in the `.bazelrc`:
```
ERROR: no such package '@rules_nixpkgs_python//': java.io.IOException: Error extracting /home/w/.cache/bazel/_bazel_w/ef63e3bfe165b0613795ad5212904c1a/external/rules_nixpkgs_python/temp15388650320627280841/v0.9.0.zip to /home/w/.cache/bazel/_bazel_w/ef63e3bfe165b0613795ad5212904c1a/external/rules_nixpkgs_python/temp15388650320627280841: Zip entries cannot refer to files outside of their directory: v0.9.0.zip has a symlink .bazelrc pointing to ../../core/.bazelrc
```

benradf (approved):
I really enjoyed reading this guide and I think it's well structured. Good work!

examples/cc-template/README.md
```diff
@@ -0,0 +1 @@
+# rules_nixpkgs guide template
```
aherrmann, line 1: Include a link to the actual guide so that people who discover this README first can find it easily.

examples/cc-template/WORKSPACE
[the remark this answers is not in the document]
aherrmann: Hmm, interesting. Does it work if you use the `.tar.gz` instead? Maybe that's just a Zip limitation?
Nonetheless, it's bad to have broken release artifacts. @benradf I think this can be fixed by cherry-picking the later changes that replace the `.bazelrc` symlinks by files with `%workspace%` relative imports. Would you have bandwidth to create a patch release to fix this?
You can use `urls = ["file:///..."]` syntax to point `http_archive` to local files, which should make local testing easier.

aherrmann (commented):
The CI failure
```
  error: getting status of '/home/runner/work/rules_nixpkgs/rules_nixpkgs/examples/cc-template/default.nix': No such file or directory
```
looks like there is a file missing.
@evertedsphere could that be a simple case of a file accidentally not checked in, or is something misconfigured and pointing at the wrong file?

guide.md
motiejus: The repository I care about right now (Uber's Go Monorepo) has `--incompatible_enable_cc_toolchain_resolution` set by default; so I presume there are more repositories "in the wild" that uses it.

Perhaps we can add a hint about it, a TODO item, and at least a link to https://github.com/tweag/rules_nixpkgs/issues/289 ?

motiejus (approved):
This is by far the most accessible overview about flakes that I've seen so far. Great job!

As an outsider, I suggest merging this as-is, and then address some points as follow-ups. A few things come to mind (that I care about):
- `--incompatible_enable_cc_toolchain_resolution` instead of `crosstool*`.
- an example `bazel build --platforms <...>` that works well with C++ and Go toolchains (or CGo!).

examples/cc-template/README.md
evertedsphere: Done.

evertedsphere:
@aherrmann @benradf I don't see a patch release in the releases list; has one been made yet? The workaround with the .tar.gz archive does indeed work, so I'm going with that for now.

examples/cc-template/WORKSPACE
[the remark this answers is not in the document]
evertedsphere: @aherrmann @benradf I don't see a patch release in the releases list; has one been made yet? The workaround with the .tar.gz archive does indeed work, so I'm going with that for now.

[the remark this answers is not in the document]
benradf: No, unfortunately I've not had time to make a patch release yet.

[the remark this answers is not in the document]
evertedsphere: Got it. I'll resolve this thread; feel free to make an issue to track this (since the zip for 0.9.0 is still broken).

examples/cc-template/WORKSPACE
```diff
@@ -0,0 +1,36 @@
+# load the http_archive rule itself
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# load rules_nixpkgs
+http_archive(
```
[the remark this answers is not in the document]
evertedsphere: Done.

examples/cc-template/WORKSPACE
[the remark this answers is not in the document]
benradf: I created #373 to track this.

benradf (approved):
Looks good!

aherrmann (commented):
Thank you! Looks great!

Issue #264, opened 2022-09-05
Split the docs generation into a separate component


avdv:
**Is your feature request related to a problem? Please describe.**

Currently, every component depends on `java`, `posix` and the `cc` toolchain (even the core). This is because of the stardoc dependency which consists of a `java_binary` and is used to generate documentation from starlark.

Ideally, the `core` should not depend on any other toolchain at all. Other toolchains should also not depend on any other toolchain that is indeed not used by it.

**Describe the solution you'd like**

Make the docs generation a component of its own.

**Describe alternatives you've considered**

avdv:
@z8v I think we should be able to get rid of the reference to the docs module in core now. It is still referenced but not actually used there.

Also, the idea was to get rid of the dependency on the docs module in all toolchains too.

z8v:
@avdv Although the docs are not used explicitly used, the core and `toolchains` reference dependencies introduced by the `docs` module. Should we introduce those dependencies directly on `core` and the `toolchains`? This means duplicating (minus the `stardoc` dependency) the `docs_dependencies_X.bzl` files on `core`.

The `toolchains` directly call `generate_documentation` on the `docs` directory. Should we remove this?

I've opened https://github.com/tweag/rules_nixpkgs/pull/304 to address the first part. Let me know if this is what you had in mind.

PR #296, opened 2022-12-21
Split the docs generation into a separate module

z8v:
Related issue: https://github.com/tweag/rules_nixpkgs/issues/264


evertedsphere:
@z8v CI should be fixed as of #297: could you rebase on `master`?

avdv:
@mergifyio rebase

avdv:
:point_up: @z8v / @evertedsphere you can have mergify rebase for you

evertedsphere:
TIL

.github/workflows/workflow.yaml
```diff
@@ -8,6 +8,14 @@ on:
...
 
 jobs:
+  cancel-running-workflows:
+    name: Cancel running workflows
+    runs-on: ubuntu-20.04
+    steps:
+      - name: cancel running workflows
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
```
avdv, line 18: Nice! Usually I tried to cancel these manually, when I was aware of them. 

But could you use `ubuntu-latest` please, so the jobs are run on the same image?

And I would prefer this to be split out into a separate commit.

core/MODULE.bazel
```diff
@@ -5,5 +5,3 @@ module(
 
 bazel_dep(name = "platforms", version = "0.0.4")
 bazel_dep(name = "bazel_skylib", version = "1.0.3")
-bazel_dep(name = "rules_sh", version = "0.3.0")
-bazel_dep(name = "stardoc", version = "0.5.1", repo_name = "io_bazel_stardoc", dev_dependency = True)
```
avdv, line 9: I am wondering if this is actually correct to remove these lines. Actually, this module still depends on these other modules since it still loads the dependencies from the new location.

avdv (changes requested):
Thank you, that is going into the right direction.

.github/workflows/workflow.yaml
z8v: Switched to `ubuntu-latest` and added a separate commit.

core/MODULE.bazel
z8v: Yes, it seems like unless the tests are a separate module we'd always need to bring those in.

docs/MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(
+    name = "docs_generation",
```
aherrmann, line 2: To not collide with other modules on the BCR this should use a more specific module name, e.g. `rules_nixpkgs_docs` to follow the pattern of the other modules.

docs/MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(
```
aherrmann, line 1: With this becoming a module I think we should also add a corresponding WORKSPACE for the legacy setup.

docs/stardoc.bzl
```diff
@@ -53,11 +53,11 @@ def stardoc(
         name,
         out,
         input,
-        aspect_template = "@rules_nixpkgs_core//docs:templates/aspect.vm",
-        func_template = "@rules_nixpkgs_core//docs:templates/func.vm",
-        header_template = "@rules_nixpkgs_core//docs:templates/header.vm",
-        provider_template = "@rules_nixpkgs_core//docs:templates/provider.vm",
-        rule_template = "@rules_nixpkgs_core//docs:templates/rule.vm",
+        aspect_template = "@io_tweag_rules_nixpkgs//docs:templates/aspect.vm",
```
aherrmann, line 56: I think these should become references to the docs module, e.g. `@rules_nixpkgs_docs//:templates/rule.vm`.

aherrmann (changes requested):
I added a couple more comments. I think we should be consistent in the module and legacy WORKSPACE worlds to avoid labels that work in only one of the worlds.

WORKSPACE
```diff
@@ -2,15 +2,20 @@ workspace(name = "io_tweag_rules_nixpkgs")
 
 # For documentation
 
+local_repository(
+    name = "io_tweag_rules_nixpkgs_docs",
```
aherrmann, line 6: With the modularization we also wanted to drop the `io_tweag_` prefix in the individual modules.
```suggestion
    name = "rules_nixpkgs_docs",
```
IIRC it was only still kept in the top-level `io_tweag_rules_nixpkgs` for backward compatibility.
All the other modules, e.g. `rules_nixpkgs_java`, don't use that prefix anymore.

core/MODULE.bazel
```diff
@@ -5,5 +5,3 @@ module(
 
 bazel_dep(name = "platforms", version = "0.0.4")
 bazel_dep(name = "bazel_skylib", version = "1.0.3")
```
aherrmann, line 7:
```suggestion
bazel_dep(name = "bazel_skylib", version = "1.0.3")
bazel_dep(name = "rules_sh", version = "0.3.0", dev_dependency = True)
```
IIUC `stardoc` is indeed no longer a dependency of `rules_nixpkgs_core`. However, `rules_sh` is still needed for the tests, but, not for regular usage. In that case it should be marked as a dev-dependency.

aherrmann (approved):
One last small comment about the rules_sh dev-dependency.
I also noticed that the docs regeneration comment in the README was out of date and pushed an update.
Apart from that this looks ready to me. Thank you for pushing this forward!

avdv (approved):
Thank you!

PR #297, opened 2022-12-22
Configure bazelisk to use Bazel 5.x

avdv:
Apparently, the Github images have bazelisk installed which downloads the latest Bazel LTS version by default.

Bazel 6 was released a few days ago and now the test-examples workflow failed for the java toolchain without nixpkgs:

```
> Running Java Toolchain Example without Nix
...
 ERROR: /home/runner/.cache/bazel/_bazel_runner/e55f8e8a4b5583b70a55184169f2761b/external/rules_java/toolchains/BUILD:212:27: no such target '@remote_java_tools//:java_compiler_jar': target 'java_compiler_jar' not declared in package '' defined by /home/runner/.cache/bazel/_bazel_runner/e55f8e8a4b5583b70a55184169f2761b/external/remote_java_tools/BUILD (Tip: use `query "@remote_java_tools//:*"` to see all the targets in that package) and referenced by '@rules_java//toolchains:toolchain_java8'
```


aherrmann (commented):
Thank you!

I think the non-Nix examples need the same configuration.

avdv:
> I think the non-Nix examples need the same configuration.

Yes. At the very least I should have changed it for the java toolchain example which was failing...

I added the `.bazeliskrc` file to each workspace to make consistent use of a single Bazel when using bazelisk.

evertedsphere (approved):

PR #298, opened 2022-12-22
cc: make CXX_FLAG configurable

### LINKED ISSUE #218: CXX_FLAGS should allow for overload

evertedsphere:
This offers an alternative to the solution proposed in #219 to the problem of `nixpkgs_cc_configure` providing a non-configurable `-x c++` in `CXX_OPTS` when calling a C compiler. We add a new optional `cc_lang` argument to `nixpkgs_cc_configure`, and then pass `-x <cc_lang>` instead.

Fixes #218.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ A dictionary from local repository name to global repository name. This allows c
 <pre>
 nixpkgs_cc_configure(<a href="#nixpkgs_cc_configure-name">name</a>, <a href="#nixpkgs_cc_configure-attribute_path">attribute_path</a>, <a href="#nixpkgs_cc_configure-nix_file">nix_file</a>, <a href="#nixpkgs_cc_configure-nix_file_content">nix_file_content</a>, <a href="#nixpkgs_cc_configure-nix_file_deps">nix_file_deps</a>, <a href="#nixpkgs_cc_configure-repositories">repositories</a>,
                      <a href="#nixpkgs_cc_configure-repository">repository</a>, <a href="#nixpkgs_cc_configure-nixopts">nixopts</a>, <a href="#nixpkgs_cc_configure-quiet">quiet</a>, <a href="#nixpkgs_cc_configure-fail_not_supported">fail_not_supported</a>, <a href="#nixpkgs_cc_configure-exec_constraints">exec_constraints</a>,
-                     <a href="#nixpkgs_cc_configure-target_constraints">target_constraints</a>, <a href="#nixpkgs_cc_configure-register">register</a>)
+                     <a href="#nixpkgs_cc_configure-target_constraints">target_constraints</a>, <a href="#nixpkgs_cc_configure-register">register</a>, <a href="#nixpkgs_cc_configure-cc_lang">cc_lang</a>)
 </pre>
 
 Use a CC toolchain from Nixpkgs. No-op if not a nix-based platform.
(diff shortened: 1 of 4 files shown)


dmadisetti:
Sure, this looks like the correct way to do things.

Plug and play into my project yields:

```
cannot find CUDA installation; provide its path via '--cuda-path', or pass '-nocudainc' to build without CUDA includes
```

While build a regular cc dependency. Not a massive blocker, I reckon this can be solved with setting environmental variables. But as in #219, auto-detection seems to work well enough so setting `none` seems to be best.

LGMT, but would suggest discussing the `none` option in the documentation since it is not well documented in clang (I had to refer to https://github.com/llvm-mirror/clang/blob/release_60/include/clang/Driver/Types.def)

aherrmann (commented):
Thanks for looking into this!

@dmadisetti does this solve the issue for you?

One issue I see with this approach is that it fixes the language for all targets built with that toolchain. Concretely, I'd expect the following type of thing to fail:
```
# WORKSPACE
nixpkgs_cc_configure(..., cc_lang = "cuda")
# BUILD
cc_binary(
    name = "cxx-bin",
    srcs = ["main.cc"],  # this is C++ code  -  I'd expect this to fail
)
cc_binary(
    name = "cude-bin",
    srcs = ["main.cu"],  # this is Cuda code
)
```

In that regard, https://github.com/bazelbuild/bazel/issues/15499 seems like a much better solution.

That said, I'm not opposed to making the `-x` flag configurable, if it solves the issue.

dmadisetti:
Yes setting `cc_lang = none` works for me!

evertedsphere:
@dmadisetti perhaps `cc_lang = none` together with `cc_library(copts = "-x whatever")` is one possible way to go about it?

aherrmann:
> Yes setting `cc_lang = none` works for me!

Nice!

> @dmadisetti perhaps `cc_lang = none` together with `cc_library(copts = "-x whatever")` is one possible way to go about it?

@evertedsphere if that works, then that would be a good thing to recommend in the docs, instead of setting `cuda` at the top-level.

dmadisetti:
It's not needed from what I can see.
However, I did some basic testing to make sure this works. It looks like the none flag is properly overloaded.

evertedsphere:
@dmadisetti what about without `none`? Does using `copts` with a vanilla `nixpkgs_cc_configure` toolchain work?

dmadisetti:
@evertedsphere no, the conflicting flags fall back to the original

dmadisetti:
LGTM

aherrmann (approved):
Nice, thank you!

PR #301, opened 2023-01-05
Fix Python container example

aherrmann:
The example was still using an older nixpkgs revision that did not provide the required Bazel version 5. This PR updates the nixpkgs revision. This PR also adds the example to at least build the image on CI, to make sure that we notice any such discrepancies in the future.


evertedsphere (approved):
LGTM

PR #304, opened 2023-01-08
Remove any docs references from core

z8v:
The `docs` module was not used directly, although its dependencies were used by core (except `stardoc`).

Related to #264 and #296

### DIFF
--- a/core/test_dependencies_2.bzl
+++ b/core/test_dependencies_2.bzl
@@ -1,0 +1,41 @@
+# NOTE: temporary for compatibility with `WORKSPACE` setup!
+# TODO: remove when migration to `bzlmod` is complete
+
+# This has to be split into `test_dependencies_1` and `test_dependencies_2`
+# because Bazel is imperative, and requires `load()` to be a top-level
+# statement.
+load("@rules_nixpkgs_core//:nixpkgs.bzl", "nixpkgs_local_repository")
+load("@rules_nixpkgs_java//:java.bzl", "nixpkgs_java_configure")
+load("@rules_sh//sh:repositories.bzl", "rules_sh_dependencies")
+load("@rules_sh//sh:posix.bzl", "sh_posix_configure")
+load("@rules_nixpkgs_posix//:posix.bzl", "nixpkgs_sh_posix_configure")
+load("@rules_nixpkgs_cc//:cc.bzl", "nixpkgs_cc_configure")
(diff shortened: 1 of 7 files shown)


core/test_dependencies_2.bzl
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    rules_sh_dependencies()
+    nixpkgs_sh_posix_configure(repository = "@nixpkgs")
+    sh_posix_configure()
+
+    nixpkgs_cc_configure(
+        name = "nixpkgs_config_cc",
+        repository = "@nixpkgs",
+    )
```
avdv, line 29: I think we don't need to configure a CC toolchain for the core module.

But the java toolchain might be required because of the `sh_test` rules. But worth a try if we can remove that too...

avdv (commented):
Yes, that's what I in mind. Maybe we can even strip down the dependencies even more.

core/test_dependencies_2.bzl
z8v: The `cc` toolchain is actually needed according to the following error:
```
../26af00e253fad871683a21acc7a6630c/external/remote_coverage_tools/BUILD:9:12: every rule of type java_binary implicitly depends upon the target '@nixpkgs
_config_cc//:toolchain', but this target could not be found because of: no such package '@nixpkgs_config_cc//': The repository '@nixpkgs_config_cc' could not be resolved: Repository '@nixpkgs_
config_cc' is not defined                 
```

avdv: I think you also need to remove the `cc` toolchain from the call to `rules_nixpkgs_dependencies` here: https://github.com/tweag/rules_nixpkgs/pull/304/files#diff-cf0d9c0f2e887febec7d80ede2216ca3d7bac9e60963a8a38bb61b983bdbca78L18

avdv: I had a quick look again, the error originates from the use of `--crosstool_top=@nixpkgs_config_cc//:toolchain` in `.bazelrc`. That setting should be removed.

Since the `.bazelrc` file in `core` is the main file that is used as a symlink target for all other components, we need to re-target the symlinks.

z8v: Indeed that was the issue. I've created a separate `.bazelrc` file for the `cc` toolchain.

avdv (approved):
Thank you!

PR #309, opened 2023-01-26
Add example for nodejs toolchain with npm_install repository rule

kczulko:
While applying latest `rules_nodejs`(`5.8.0`) to one of the projects I'm working on, I saw there's no longer an option to pass `vendored_node` to `npm_install` rule.

Following [this "migration guide"](https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0) it was found out (thx @avdv @aherrmann ) that using proper `npm_install::node_reporitory` value does the job.

However there's an issue of suffixing toolchain names which rules_nodejs does - see `nixpkgs-nodejs_linux_amd64` as one of the names. E.g. now macos build fails with:
```
Error in path: Unable to load package for @nixpkgs-nodejs_darwin_amd64//:bin/node: The repository '@nixpkgs-nodejs_darwin_amd64' could not be resolved: Repository '@nixpkgs-nodejs_darwin_amd64' is not defined
```

Probably we have to mimic the behavior that [rules_haskell does](https://github.com/tweag/rules_haskell/blob/8e0189dc8b444f81ff33e5f69ae4489cb342bab6/haskell/asterius/asterius_dependencies.bzl#L96-L113) (thx @ylecornec )

### DIFF
--- a/examples/toolchains/nodejs/WORKSPACE
+++ b/examples/toolchains/nodejs/WORKSPACE
@@ -28,9 +28,9 @@ nixpkgs_git_repository(
     sha256 = "0f8c25433a6611fa5664797cd049c80faefec91575718794c701f3b033f2db01",
 )
 
-load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_nodejs_configure")
-nixpkgs_nodejs_configure(
-  name = "nixpkgs-nodejs_linux_amd64",
+load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_nodejs_configure_platforms")
+nixpkgs_nodejs_configure_platforms(
+  name = "nixpkgs-nodejs",
   repository = "@nixpkgs",
 )
 
--- a/examples/toolchains/nodejs/WORKSPACE
+++ b/examples/toolchains/nodejs/WORKSPACE
@@ -7,15 +7,9 @@ local_repository(
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
-    name = "rules_nodejs",
-    sha256 = "5aef09ed3279aa01d5c928e3beb248f9ad32dde6aafe6373a8c994c3ce643064",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.5.3/rules_nodejs-core-5.5.3.tar.gz"],
-)
-
-http_archive(
     name = "build_bazel_rules_nodejs",
-    sha256 = "f10a3a12894fc3c9bf578ee5a5691769f6805c4be84359681a785a0c12e8d2b6",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.5.3/rules_nodejs-5.5.3.tar.gz"],
+    sha256 = "dcc55f810142b6cf46a44d0180a5a7fb923c04a5061e2e8d8eb05ccccc60864b",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-5.8.0.tar.gz"],
 )
 
 load("@build_bazel_rules_nodejs//:repositories.bzl", "build_bazel_rules_nodejs_dependencies")
@@ -35,11 +29,17 @@ nixpkgs_git_repository(
 )
 
 load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_nodejs_configure")
-
 nixpkgs_nodejs_configure(
+  name = "nixpkgs-nodejs_linux_amd64",
   repository = "@nixpkgs",
 )
 
-load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories", "yarn_install")
+load("@build_bazel_rules_nodejs//:index.bzl", "npm_install")
+npm_install(
+    name = "npm",
+    exports_directories_only = True,
+    node_repository = "nixpkgs-nodejs",
+    package_json = "//:package.json",
+    package_lock_json = "//:package-lock.json",
+)
 
-node_repositories()
--- a/examples/toolchains/nodejs/shell.nix
+++ b/examples/toolchains/nodejs/shell.nix
@@ -1,3 +1,3 @@
 { pkgs ? import ./nixpkgs.nix { } }:
 
-pkgs.mkShellNoCC { nativeBuildInputs = [ pkgs.bazel_6 ]; }
+pkgs.mkShellNoCC { nativeBuildInputs = with pkgs; [ bazel_6 nodejs ]; }
--- a/examples/toolchains/nodejs/shell.nix
+++ b/examples/toolchains/nodejs/shell.nix
@@ -1,3 +1,3 @@
 { pkgs ? import ./nixpkgs.nix { } }:
 
-pkgs.mkShellNoCC { nativeBuildInputs = with pkgs; [ bazel_6 nodejs ]; }
+pkgs.mkShellNoCC { nativeBuildInputs = with pkgs; [ bazel_6 ]; }
--- a/toolchains/nodejs/BUILD.bazel
+++ b/toolchains/nodejs/BUILD.bazel
@@ -13,5 +13,6 @@ bzl_library(
     srcs = ["//:nodejs.bzl"],
     deps = [
         "@rules_nixpkgs_core//:nixpkgs",
+        "@rules_nodejs//nodejs:bzl",
     ],
 )
--- a/toolchains/nodejs/WORKSPACE
+++ b/toolchains/nodejs/WORKSPACE
@@ -31,8 +31,14 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "build_bazel_rules_nodejs",
-    sha256 = "f10a3a12894fc3c9bf578ee5a5691769f6805c4be84359681a785a0c12e8d2b6",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.5.3/rules_nodejs-5.5.3.tar.gz"],
+    sha256 = "dcc55f810142b6cf46a44d0180a5a7fb923c04a5061e2e8d8eb05ccccc60864b",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-5.8.0.tar.gz"],
+)
+
+http_archive(
+    name = "rules_nodejs",
+    sha256 = "08337d4fffc78f7fe648a93be12ea2fc4e8eb9795a4e6aa48595b66b34555626",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-core-5.8.0.tar.gz"],
 )
 
 load("//:nodejs.bzl", "nixpkgs_nodejs_configure")
--- a/toolchains/nodejs/nodejs.bzl
+++ b/toolchains/nodejs/nodejs.bzl
@@ -2,14 +2,23 @@ load("@rules_nixpkgs_core//:nixpkgs.bzl", "nixpkgs_package")
 load("@rules_nixpkgs_core//:util.bzl", "ensure_constraints")
 load("@rules_nodejs//nodejs/private:toolchains_repo.bzl", "PLATFORMS")
 
+
+def _mk_mapping(rules_nodejs_platform_name):
+    constraints = PLATFORMS[rules_nodejs_platform_name].compatible_with
+    return struct(
+        rules_nodejs_platform = rules_nodejs_platform_name,
+        exec_constraints = constraints,
+        target_constraints = constraints,
+    )
+
 # obtained (and matched) from:
 # nixpkgs search: https://search.nixos.org/packages?channel=22.11&show=nodejs&from=0&size=50&sort=relevance&type=packages&query=nodejs
 # rules_nodejs: https://github.com/bazelbuild/rules_nodejs/blob/a5755eb458c2dd8e0e2cf9b92d8304d9e77ea117/nodejs/private/toolchains_repo.bzl#L20
-_nodejs_nix_to_bazel_platforms_map = {
-    "aarch64-darwin": "darwin_arm64",
-    "x86_64-linux": "linux_amd64",
-    "x86_64-darwin": "darwin_amd64",
-    "aarch64-linux": "linux_arm64",
+DEFAULT_PLATFORMS_MAPPING = {
+  "aarch64-darwin": _mk_mapping("darwin_arm64"),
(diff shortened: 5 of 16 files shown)


examples/toolchains/nodejs/WORKSPACE
```diff
@@ -35,11 +29,17 @@ nixpkgs_git_repository(
 )
 
 load("@io_tweag_rules_nixpkgs//nixpkgs:nixpkgs.bzl", "nixpkgs_nodejs_configure")
-
 nixpkgs_nodejs_configure(
+  name = "nixpkgs-nodejs_linux_amd64",
```
aherrmann, line 33: As you point out, a possible fix for this may be to mimic [rules_haskell's behavior](https://github.com/tweag/rules_haskell/blob/8e0189dc8b444f81ff33e5f69ae4489cb342bab6/haskell/asterius/asterius_dependencies.bzl#L96-L113) within `nixpkgs_nodejs_configure`, or perhaps better a `nixpkgs_nodejs_configure_platforms` which invokes `nixpkgs_nodejs_configure` once for each platform.

`nixpkgs_nodejs_configure` could perhaps be changed along the lines of this (untested):
```diff
diff --git a/toolchains/nodejs/nodejs.bzl b/toolchains/nodejs/nodejs.bzl
index d4cef6b..4c20c95 100644
--- a/toolchains/nodejs/nodejs.bzl
+++ b/toolchains/nodejs/nodejs.bzl
@@ -3,7 +3,7 @@ load("@rules_nixpkgs_core//:util.bzl", "ensure_constraints")

 _nodejs_nix_content = """\
 let
-    pkgs = import <nixpkgs> {{ config = {{}}; overlays = []; }};
+    pkgs = import <nixpkgs> {{ config = {{}}; overlays = []; system = {system}; }};
     nodejs = pkgs.{attribute_path};
 in
 pkgs.buildEnv {{
@@ -74,6 +74,7 @@ def nixpkgs_nodejs_configure(
   nixopts = [],
   fail_not_supported = True,
   quiet = False,
+  system = None,
   exec_constraints = None,
   target_constraints = None,
 ):
@@ -83,6 +84,7 @@ def nixpkgs_nodejs_configure(
     if not nix_file and not nix_file_content:
       nix_file_content = _nodejs_nix_content.format(
         attribute_path = attribute_path,
+        system = repr(system) if system else "builtins.currentSystem",
       )

     nixpkgs_package(
```
Callers that set `system` must also set `exec|target_constraints` accordingly.

examples/toolchains/nodejs/WORKSPACE
```diff
@@ -35,11 +29,17 @@ nixpkgs_git_repository(
...
   repository = "@nixpkgs",
 )
 
-load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories", "yarn_install")
+load("@build_bazel_rules_nodejs//:index.bzl", "npm_install")
+npm_install(
+    name = "npm",
+    exports_directories_only = True,
+    node_repository = "nixpkgs-nodejs",
```
aherrmann, line 41: The example is designed to test both a with-nix and a without-nix configuration. In the latter case this should use the default bindist `node_repository`, so should be set to `None`.

aherrmann (commented):
@kczulko Thank you for raising and investigating this!

> Probably we have to mimic the behavior that [rules_haskell does](https://github.com/tweag/rules_nixpkgs/pull/309) (thx @ylecornec )

The link points to this PR again. [Did you mean this](https://github.com/tweag/rules_haskell/blob/8e0189dc8b444f81ff33e5f69ae4489cb342bab6/haskell/asterius/asterius_dependencies.bzl#L96-L113)?

examples/toolchains/nodejs/WORKSPACE
kczulko: Why do we need this (additional `system` param)? I probably know it's not targeted to remote build execution, but isn't enough just to take all the platforms that rules_nodejs defines (yes... I know it's pointless e.g. in case of windows) and iterate over them (like I did it in my next commit)?

aherrmann: > isn't enough just to take [all the platforms that rules_nodejs defines](https://github.com/bazelbuild/rules_nodejs/blob/a5755eb458c2dd8e0e2cf9b92d8304d9e77ea117/nodejs/private/toolchains_repo.bzl#L20) [...] and iterate through them

In that case the platform in the workspace name does not match the platform for which the contained nix-import works. E.g. if you build on x64 Linux `nixpkgs-nodejs_macos_arm64` would still give you an x64 Linux nodejs installation, because `system` defaults to `builtins.currentSystem`, so when this is evaluated on x64 Linux, then that's what you get.

It works for local builds in the case of it being a workaround to the naming requirement. But, it's pretty misleading. (For that route it would be better to stick to a single `nixpkgs_nodejs_configure` and select the name based on the host platform. Or, perhaps use the [implicit `repo_mapping` attribute](https://bazel.build/reference/be/workspace#local_repository.repo_mapping) on `npm_install` - if that works.)

kczulko: Ok, so I have next question.

When you evaluate your build on x86-64_linux, why you want to get nodejs for e.g. macos (or any other architecture)? Maybe it's a general question, why one evaluating build on architecture A, would like to execute (on architecture A) the binary compiled for architecture B? Why in this case (nodejs) we **don't** want to have 1:1 matching between evaluation architecture and node binary "version".

Sorry - it sounds elementary, but this answer would give me better understanding of what we're doing here.

aherrmann: > When you evaluate your build on x86_64-linux, why you want to get nodejs for e.g. macos (or any other architecture)?

That would only be relevant in a remote execution setting, you can find more details [here](https://github.com/tweag/rules_nixpkgs/issues/180#issuecomment-1284656149) and [here](https://github.com/tweag/rules_nixpkgs/issues/180#issuecomment-1381090829). In brief, if you evaluate on a Linux host to build on a MacOS remote executor, then you'd need the MacOS nodejs.

However, as described in https://github.com/tweag/rules_nixpkgs/issues/180, remote execution is not well supported by rules_nixpkgs, yet. So, we don't need to make support for it a blocker for this issue.

My point is more about not generating misleading external workspaces. The problem with the [approach here](https://github.com/tweag/rules_nixpkgs/pull/309/commits/814b9c45c7face4c590c2b9bfd12de8e74c666fc#diff-150d5ce7d9b192ca069368c968ea722a87abd8ca74b98149f0b54e1950c9d60eR115) is that it generates multiple copies of the same external workspace under different names such that on an x64 Linux host both, `nixpkgs_nodejs_linux_amd64` and `nixpkgs_nodejs_macos_arm64`, generate an x64 Linux nodejs. That is bound to cause confusion down the line.

So, if we generate multiple workspaces, then they should be correct, i.e. `nixpkgs_nodejs_linux_amd64` always gives you an x64 Linux nodejs, and `nixpkgs_nodejs_macos_arm64` always gives you an arm64 MacOS nodejs. Or, alternatively, we only generate a workspace for the current host.

kczulko: Thank you for this answer. I tried to address this issue (probably in a weak way). I am fixing `target|exec_constraints` to the ones that are defined under `rules_nodejs`. Please take a look.

toolchains/nodejs/BUILD.bazel
```diff
@@ -13,5 +13,6 @@ bzl_library(
     srcs = ["//:nodejs.bzl"],
     deps = [
         "@rules_nixpkgs_core//:nixpkgs",
+        "@rules_nodejs//nodejs:bzl",
```
kczulko, line 16: Is it ok to add a dependency here?

toolchains/nodejs/nodejs.bzl
```diff
@@ -105,3 +119,18 @@ def nixpkgs_nodejs_configure(
...
+):
+    for nix_platform, bazel_platform in _nodejs_nix_to_bazel_platforms_map.items():
+        # get the constaints defined under rules_nodejs PLATFORMS variable
+        constraints = PLATFORMS[bazel_platform].compatible_with
+        nixpkgs_nodejs_configure(
+            name = "{}_{}".format(name, bazel_platform),
+            nix_platform = nix_platform,
+            exec_constraints = constraints,
+            target_constraints = constraints,
+            **kwargs,
```
kczulko, line 135: @aherrmann 
I think I know what I did wrong. So in general this ^^ code can be the default path for a user. However, we should allow users to setup whatever `platform/nix_system` they want. This platforms intersection (`__nodejs_nix_to_bazel_platforms_map`) is also something that should land in the parameters list. Maybe it should be some kind of dictionary/struct with the following "schema":
```python
{
   "x86_64-linux": {
      "rules_nodejs_platform": "linux_amd64",
      "exec_constraints": [
         ...
      ],
      "target_constraints": [
         ...
      ]
   }
   "aarch64-linux": {
      "rules_nodejs_platform": "linux_arm64",
      "exec_constraints": [
         ...
      ],
      "target_constraints": [
         ...
      ]
   }
}
```
I mean it could be something similar to [PLATFORMS from rules_nodejs](https://github.com/bazelbuild/rules_nodejs/blob/stable/nodejs/private/toolchains_repo.bzl#L20) but it should include nixpkgs `system` value. By default, rules_nixpkgs code can generate configurations for the intersection that I've presented but it should also allow users to define their own setups (whatever it can be...) as we allow to define custom `nix_file` (which can contain anything).

toolchains/nodejs/WORKSPACE
```diff
@@ -31,8 +31,14 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
...
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.5.3/rules_nodejs-5.5.3.tar.gz"],
+    sha256 = "dcc55f810142b6cf46a44d0180a5a7fb923c04a5061e2e8d8eb05ccccc60864b",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-5.8.0.tar.gz"],
+)
+
+http_archive(
+    name = "rules_nodejs",
+    sha256 = "08337d4fffc78f7fe648a93be12ea2fc4e8eb9795a4e6aa48595b66b34555626",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-core-5.8.0.tar.gz"],
 )
```
aherrmann, line 42: [CI fails with not up-to-date docs](https://github.com/tweag/rules_nixpkgs/actions/runs/4024688026/jobs/6917002150#step:5:1086)
And running
```
bazel run //docs:update-README.md
```
fails with 
```
The following chain of repository dependencies lead to the missing definition.
 - @rules_nodejs
This could either mean you have to add the '@rules_nodejs' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
```
Indeed, `@rules_nodejs` is only listed in `toolchains/nodejs/WORKSPACE` but not imported in the top-level workspace. @avdv what's the best place to import that at this point, taking the factoring out of the docs workspace into account?

@kczulko is this actually needed? As I understand the `rules_nodejs` [setup instructions](https://github.com/bazelbuild/rules_nodejs/releases/tag/5.8.0) this is an optional thing if one wants only the core, or is it mandatory?

kczulko: > @kczulko is this actually needed? As I understand the rules_nodejs [setup instructions](https://github.com/bazelbuild/rules_nodejs/releases/tag/5.8.0) this is an optional thing if one wants only the core, or is it mandatory?

Is this (rules_nodejs (core)) really needed? Maybe not. However, now I am using its PLATFORMS "enumeration" to do the mapping between nixpkgs <-> rules_nodejs platforms. This as well can be ~copy-pasted to rules_nixpkgs repo - just wanted to avoid boilerplate duplication.

aherrmann: I see, I thought the core package would be a subset, but the platforms dict is only included in core. Okay, in that case both will have to be added as dependencies.

avdv: With the factoring out of the docs workspace, one should just run `bazel run //docs:update-README.md` inside the appropriate toolchain directory.

kczulko:
@aherrmann @avdv 
I think it's ready for the next review round. I think I am also missing some docs about this new way of configuring rules_nodejs under nix. I added rules_nodejs_core as explicit dependency of this project. Otherwise there were some "circular" (?) dependency issues which lead to a failures.

toolchains/nodejs/nodejs.bzl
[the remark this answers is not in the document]
aherrmann: Naming and forwarding them explicitly is preferable for Stardoc API documentation generation, as it lets you document the parameters in the functions doc string, see [here for example](https://github.com/tweag/rules_nixpkgs/pull/315/files#diff-5d11a3405724b82536cba9052da5603190c6ddb3c5ff433673638b028fe75521R73).

examples/toolchains/nodejs/WORKSPACE
aherrmann: Oh, I see it's [currently not tested without Nix](https://github.com/tweag/rules_nixpkgs/blob/dfdc4c7c786822c57a11614f7a066fa042b3051d/.github/workflows/workflow.yaml#L93). Sorry, misread that before. In that case we don't need to worry about this.

toolchains/nodejs/BUILD.bazel
aherrmann: Yes, necessary in fact.

aherrmann (commented):
Thank you! That looks good!

> Otherwise there were some "circular" (?) dependency issues which lead to a failures.

Bazel reports missing external repositories like that sometimes.

> I think I am also missing some docs about this new way of configuring rules_nodejs under nix.

A good place to add them would be the docstring of the new `nixpkgs_nodejs_configure_platforms` and also a docstring in `nixpkgs_nodejs_configure` pointing to it. (Also a note in the CHANGELOG). You can look at other macros for the format of docstrings.

examples/toolchains/nodejs/shell.nix
```diff
@@ -1,3 +1,3 @@
 { pkgs ? import ./nixpkgs.nix { } }:
 
-pkgs.mkShellNoCC { nativeBuildInputs = [ pkgs.bazel_6 ]; }
+pkgs.mkShellNoCC { nativeBuildInputs = with pkgs; [ bazel_6 nodejs ]; }
```
aherrmann, line 3:
```suggestion
pkgs.mkShellNoCC { nativeBuildInputs = with pkgs; [ bazel_6 ]; }
```
Adding `nodejs` to the shell is not necessary for the example to build thanks to the `node_repository` attribute to `npm_install`. It works without in a pure Nix shell.

aherrmann (approved):
Looks great! Thank @kczulko for pushing this through, great work!
Only one small comment. Feel free to apply and then add the `merge-queue` label.

PR #313, opened 2023-02-01
Update to Bazel version 6 and nixpkgs 22.11

aherrmann:
Update nixpkgs to 22.11 and Bazel to version 6 in the top-level workspace and the example workspaces.

Also sets the missing sha256 on the rules_go import in toolchains/go.


avdv (approved):
Nice, thank you!

PR #314, opened 2023-02-01
Update the Rust toolchain example

mrkkrp:
Update `rules_rust` to 0.17.0.


examples/toolchains/rust/nixpkgs.json
```diff
@@ -2,6 +2,6 @@
   "owner": "NixOS",
   "repo": "nixpkgs",
   "branch": "22.05",
-  "rev": "ce6aa13369b667ac2542593170993504932eb836",
-  "sha256": "0d643wp3l77hv2pmg2fi7vyxn4rwy0iyr8djcw1h5x72315ck9ik"
+  "rev": "73247c5ab2bc2037d7dd0d310f1fb43ed191edad",
+  "sha256": "1213j2gvsmzmhpyi2jyn729s9d0gbn56k3d873iknd5ilb2rvpcr"
```
aherrmann, line 6: What channel is that from? It doesn't look like it's 22.05 anymore. Would 22.11 suffice (as being updated [here](https://github.com/tweag/rules_nixpkgs/pull/313/files#diff-e539c550d3ff8868701fc7fc237c08a365c969ec8561a6104205c9bfd0d4e278)), or does it have to be more recent? If so, could you update the branch field accordingly?

examples/toolchains/rust/patches/001-rules-rust-crate-universe.patch
```diff
@@ -24,36 +24,33 @@ index 28f6d14..a2eeb3a 100644
...
++        rust_toolchain_cargo_template = "@rust_{system}_{arch}//:bin/{tool}",
++        rust_toolchain_rustc_template = "@rust_{system}_{arch}//:bin/{tool}",
++):
      """Define dependencies of the `cargo-bazel` Rust target
  
      Args:
-         rust_version (str, optional): The version of rust to use when generating dependencies.
-         bootstrap (bool, optional): If true, a `cargo_bootstrap_repository` target will be generated.
-+        rust_toolchain_cargo_template (str, optional): The template to use for finding the host `cargo` binary.
-+        rust_toolchain_rustc_template (str, optional): The template to use for finding the host `rustc` binary.
```
aherrmann, line 47: why remove those from the patch?

mrkkrp: This was not intentional. The idea was to get it working, this patch being for internal use, not to be included upstream. I'm going to restore this and add the lines in another place where the API is also extended (`cargo_bazel_bootstrap`, which did not have such lines in the old patch).

aherrmann (approved):
Thank you!

PR #315, opened 2023-02-03
Pass the workspace name as an attribute

aherrmann:
This is needed for bzlmod compatibility where the `repository_ctx` only exposes the mangled name (e.g. `_main~non_module_deps~nixpkgs`), but `nixpkgs_local_repository` and others require the unmangled name (e.g. `nixpkgs`).

repository rules are converted into macros so that users don't need to manually specify the `unmangled_name` attribute.


avdv (approved):
LGTM, thanks!

PR #320, opened 2023-02-09
Separate toolchains/java tests and run with bzlmod enabled

aherrmann:
Depends on #319

Creates a new module in `testing/java` and moves the tests from `toolchains/java` into this module following the motivation from https://github.com/tweag/rules_sh/issues/41.

Makes the Java toolchain registration optional, as the corresponding `native.register_toolchains` cannot be called in a repository macro under bzlmod.

Building `toolchains/java` with bzlmod enabled raised a problem: The module depends on `rules_nixpkgs_core`, but cannot find it in the Bazel Central Registry. The obvious solution, [`local_path_override`](https://bazel.build/rules/lib/globals#local_path_override) does not work, because that is only allowed in the main module, but `rules_nixpkgs_java` is a library module. The next best solution [`--override_module`](https://bazel.build/reference/command-line-reference#flag--override_module) is awkward, because that flag requires an absolute path, does not support `%workspace%` expansion, and must not be set for the module it is overriding.

The solution chosen in this PR is to use a [local Bazel registry](https://bazel.build/external/registry). The corresponding configuration flag `--registry` does support `%workspace%` expansion and is therefore easier to configure.


aherrmann:
The correct configuration for dpulls is to have this PR based on the dependency branch. However, due to the CI configuration this doesn't run CI. So I'm switch the base branch back to master. Perhaps we should change the CI configuration to allow CI on all PRs.

registry/README.md
```diff
@@ -0,0 +1,10 @@
+# Local Bazel Registry for rules\_nixpkgs modules
+
+This [local Bazel registry][registry] is only intended for testing and local
+development purposes. The rules\_nixpkgs is split into multiple Bazel modules
```
benradf, line 4: Missing a word? Maybe:
> The rules\_nixpkgs repository is split into ...

benradf (approved):
Looks good :+1:

registry/README.md
aherrmann: Good catch! Fixed.

PR #329, opened 2023-02-16
Separate testing module and bzlmod tests for toolchains/rust

aherrmann:
Creates a new module in `testing/rust` and adds tests for `toolchains/rust` into this module following the motivation from https://github.com/tweag/rules_sh/issues/41.

`toolchains/rust` didn't have any tests before. This adds a simple `rust_binary` target, and a skylib-unittest that confirms that the resolved Rust toolchain is indeed the one provided by rules\_nixpkgs.

rules\_rust itself does not have a bzlmod module definition, yet, and is not yet listed on the Bazel Central Registry. To enable testing in bzlmod already anway, this adds an entry for rules\_rust to the local registry. Notably, this entry is minimal and only captures what is required for the tests under `toolchains/rust`. It does not constitute a full Bazel module definition for rules\_rust.

This also makes the Rust toolchain registration optional, as it is not allowed to invoke `native.register_toolchains` outside of `MODULE.bazel` in bzlmod mode.


benradf (approved):
Looks good :+1:

PR #332, opened 2023-02-17
Separate testing module and bzlmod tests for toolchains/nodejs

aherrmann:
Creates a new module in `testing/nodejs` and moves the tests for `toolchains/nodejs` into this module following the motivation from https://github.com/tweag/rules_sh/issues/41.

This adds a skylib-unittest that confirms that the resolved NodeJS toolchain is indeed the one provided by rules\_nixpkgs.

The nodejs rules are split into the core module rules\_nodejs, which is also listed on the Bazel Central Registry, and the deprecated build\_bazel\_rules\_nodejs which does not have a Bazel module and is to be replaced by rules\_js going forward. For `testing/nodejs` we add the deprecated build\_bazel\_rules\_nodejs to `WORKSPACE.bzlmod` to still import it in bzlmod mode.

This also makes the NodeJS toolchain registration optional, as it is not allowed to invoke `native.register_toolchains` outside of `MODULE.bazel` in bzlmod mode.


benradf (approved):
Looks good!

testing/nodejs/WORKSPACE.bzlmod
```diff
@@ -0,0 +1,14 @@
...
+# we cannot expect a Bazel module to be created for it.
+
+# TODO[AH] Migrate rules_nixpkgs_nodejs to use rules_js for any JavaScript
+# targets, rules_nodejs can still be used for the core functionality, the
+# toolchain in particular.
+http_archive(
+    name = "build_bazel_rules_nodejs",
+    sha256 = "dcc55f810142b6cf46a44d0180a5a7fb923c04a5061e2e8d8eb05ccccc60864b",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-5.8.0.tar.gz"],
+)
```
kczulko, line 14: @aherrmann 
AFAIR this dependency is used only to obtain a single variable (I think it was `PLATFORMS` or something). One option to get rid of this is to hardcode those platforms that we're interested in. Downside of this solution is the need of keeping it synchronized (without the a priori knowledge that something got changed). For it's ~ok to accept it :)

aherrmann: `PLATFORMS` [comes from `@rules_nodejs`](https://github.com/tweag/rules_nixpkgs/blob/master/toolchains/nodejs/nodejs.bzl#L3), which corresponds to `https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.0/rules_nodejs-core-5.8.0.tar.gz`. That's separate from `build_bazel_rules_nodejs`. This dependency here is needed for the [`nodejs_test` target](https://github.com/tweag/rules_nixpkgs/pull/332/files#diff-4bd2f834ad36fd326dedd9619e1d7873e0542173559c844809b46e7cf761990dR6) under `testing/nodejs/tests`.

Issue #328, opened 2023-02-14
`nixpkgs_cc_configure` should allow `attribute_path` without having a nix file


evertedsphere:
**Is your feature request related to a problem? Please describe.**
It should be possible to write
```python
nixpkgs_cc_configure(
  repository = "@nixpkgs",
  nix_file_content = "(import <nixpkgs> {}).gcc11",
  name = "nixpkgs_config_cc",
)
``` 
as
```python
nixpkgs_cc_configure(
  repository = "@nixpkgs",
  attribute_path = "gcc11",
  name = "nixpkgs_config_cc",
)
```

**Describe the solution you'd like**
The `attribute_path` should be drawn from the `repository` if neither of `nix_file{,_content}` is given.

**Describe alternatives you've considered**

**Additional context**

z8v:
Closing this, since #334 was merged.

PR #334, opened 2023-02-18
Update `nixpkgs_cc_configure` to work without `nix_file` or `nix_content`

z8v:
The simple example for the cc toolchain was updated to showcase the usage. I'm not sure if this should have a separate test altogether.

Related issue: #328

### DIFF
--- a/toolchains/cc/cc.bzl
+++ b/toolchains/cc/cc.bzl
@@ -342,6 +342,13 @@ def nixpkgs_cc_configure(
     )
     ```
     ```
+    # alternate usage without specifying `nix_file` or `nix_file_content`
+    nixpkgs_cc_configure(
+      repository = "@nixpkgs",
+      attribute_path = "gcc11",
+    )
+    ```
+    ```
     # use the `stdenv.cc` compiler (the default of the given @nixpkgs repository)
     nixpkgs_cc_configure(
       repository = "@nixpkgs",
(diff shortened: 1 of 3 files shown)


evertedsphere (approved):
Looks good!

toolchains/cc/cc.bzl
```diff
@@ -383,7 +390,14 @@ def nixpkgs_cc_configure(
         nix_expr = nix_file_content
 
     if attribute_path and nix_expr == None:
-        fail("'attribute_path' requires one of 'nix_file' or 'nix_file_content'", "attribute_path")
+        nixopts.extend([
+            "--argstr",
+            "ccType",
+            "ccTypeExpression",
+            "--arg",
```
benradf, line 397: Perhaps this would be a bit simpler?
```
@@ -388,17 +388,11 @@ def nixpkgs_cc_configure(
         nix_file_deps.append(nix_file)
     elif nix_file_content:
         nix_expr = nix_file_content
-
-    if attribute_path and nix_expr == None:
-        nixopts.extend([
-            "--argstr",
-            "ccType",
-            "ccTypeExpression",
-            "--arg",
-            "ccExpr",
-            "(import <nixpkgs> {{}}).{0}".format(attribute_path),
-        ])
     elif attribute_path:
+        nix_expr = "(import <nixpkgs> {{}}).{0}".format(attribute_path)
+        attribute_path = None
+
+    if attribute_path:
         nixopts.extend([
             "--argstr",
             "ccType",


```
I think it's equivalent but the logic is a little convoluted so I may be wrong.

benradf (approved):
I think this is a good change. Just one suggestion about the logic around `nix_expr` and `attribute_path`.

toolchains/cc/cc.bzl
benradf: Not sure how to make a GitHub suggested change that I've seen people do sometimes. Here's an actual commit though:
https://github.com/tweag/rules_nixpkgs/commit/841908b4288481a0a022193d76bb251170212d1e

PR #338, opened 2023-02-22
Generate the local nixpkgs repository from flake.lock

### LINKED ISSUE #327: Add a `flake_lockfile` option to `nixpkgs_local_configure`
Currently, when using `rules_nixpkgs` in a flake-enabled project, defining a `nixpkgs_local_repository` that is consistent with the nixpkgs in the flake requires the use of a `nixpkgs.nix` shim of the following form:
```nix
let
  lock = builtins.fromJSON (builtins.readFile ./flake.lock);
  spec = lock.nodes.nixpkgs.locked;
  nixpkgs = fetchTarball "https://github.com/${spec.owner}/${spec.repo}/archive/${spec.rev}.tar.gz";
in
import nixpkgs
```

We could allow the user to specify a `flake.lock` file and perform this shimming ourselves. This would reduce friction in simple use-cases, where we risk losing prospective users who are new to Nix and do not yet have much experience translating between "old" and "new" Nix.

**Additional context**

z8v:
This will remove the need for extra steps to convert from a flakes based project.

An example was added that showcases the `nix_flake_lock_file` argument on `nixpkgs_local_repository`.

fixes #327


core/nixpkgs.bzl
```diff
@@ -106,18 +106,40 @@ def nixpkgs_git_repository(
...
         repository_ctx.file(
             target,
             content = repository_ctx.attr.nix_file_content,
             executable = False,
         )
-    else:
+    elif repository_ctx.attr.nix_file:
         target = cp(repository_ctx, repository_ctx.attr.nix_file)
+    else:
+        lock_filename = str(repository_ctx.attr.nix_flake_lock_file).split(":")[1]
```
z8v, line 124: Maybe there is a better way to get the filename?

aherrmann: You can follow the `cp` pattern above, I think.
```
lock_filename = cp(repository_ctx, repository_ctx.attr.nix_flake_lock_file).
```

core/nixpkgs.bzl
```diff
@@ -106,18 +106,40 @@ def nixpkgs_git_repository(
...
         target = "default.nix"
         repository_ctx.file(
             target,
             content = repository_ctx.attr.nix_file_content,
             executable = False,
         )
-    else:
+    elif repository_ctx.attr.nix_file:
         target = cp(repository_ctx, repository_ctx.attr.nix_file)
+    else:
```
benradf, line 123: Minor, but it looks like the `if` statement above doesn't catch the case where all of `nix_file`, `nix_file_content` and `nix_flake_lock_file` are `None`? Maybe this `else` could be `elif repository_ctx.attr.nix_flake_lock_file`, with an `else` at the end to `fail` when none of them are set.

benradf (approved):
One comment, but otherwise looks good :+1:

PR #343, opened 2023-03-08
Initial implementation of python modules with poetry

layus:
This has been long in the making, and currently it only works in a very
restricted setup. Nonetheless, it comes with a working example, so comments are
welcome.


testing/python-poetry/tests/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+load("@nixpkgs_python_toolchain//:defs.bzl", "interpreter")
+load("@nix_pip//:requirements.bzl", "requirement")
```
aherrmann, line 2: Should it perhaps be called `nix_poetry` instead? Since it's based on poetry instead of pip...

Apart from that I like the similarity to the [rules_python pip API](https://github.com/bazelbuild/rules_python#installing-third_party-packages). That should help make adoption easy.

testing/python-poetry/tests/nixpkgs_repositories.bzl
```diff
@@ -0,0 +1,53 @@
...
+
+    nixpkgs_python_repository(
+        name = "nix_pip",
+        repository = "@nixpkgs",
+        nix_file = "//:poetry.nix",
+        nix_file_deps = [
+            "//:pyproject.toml",
+            "//:poetry.lock",
+        ],
+    )
```
aherrmann, line 52: Apart from this bit the file looks identical to `testing/python`. So, by that token it looks like they could be merged.

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+        quiet = False,
+        ):
+    """Define a collection of python modules.
+
+    Args:
+      name: The name for the created module set.
+      repository: See [`nixpkgs_package`](#nixpkgs_package-repository).
+      repositories: See [`nixpkgs_package`](#nixpkgs_package-repositories).
+      nix_file: See [`nixpkgs_package`](#nixpkgs_package-nix_file).
+      nix_file_deps: See [`nixpkgs_package`](#nixpkgs_package-nix_file_deps).
```
aherrmann, line 318:
```suggestion
      nix_file_deps: See [`nixpkgs_package`](#nixpkgs_package-nix_file_deps).
      quiet: See [`nixpkgs_package`](#nixpkgs_package-quiet).
```
Missing arg?

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+    """
+
+    generated_deps_name = "_generated_{}_deps".format(name)
+
+    nixpkgs_package(
+        name = generated_deps_name,
+        nix_file_content = """
```
aherrmann, line 325: Since the question came up in the Q/A. You can [see here](https://github.com/tweag/rules_nixpkgs/blob/3748711f010bd191b0867344d615f9dc98b6eaf9/toolchains/cc/cc.bzl#L429-L430) how this can be done with a `nix_file` attribute and a file stored in `rules_nixpkgs_python`.

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+{ nix_file }:
```
aherrmann, line 326: In the [CC case](https://github.com/tweag/rules_nixpkgs/blob/3748711f010bd191b0867344d615f9dc98b6eaf9/toolchains/cc/cc.nix#L5-L10) we support `nix_file` as well as `nix_file_content` for API consistency. However, in this case this is the file generated by poetry2nix, correct? So, only supporting `nix_file` seems reasonable.

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+  nixpkgs = import <nixpkgs> {};
+  pythonExpr = import nix_file;
+  inherit (pythonExpr) python pkgs;
+
+  isPythonModule = drv: drv ? pythonModule && drv ? pythonPath;
+  filterPythonModules = builtins.filter isPythonModule;
+
+  # Ensure the dependency list is unique, otherwise bazel complains about
+  # duplicate names in the generated python_module() rule
+  unique = list: builtins.attrNames (builtins.listToAttrs (builtins.map (x: {
```
aherrmann, line 337: Isn't there a `uniq` function in nixpkgs' `lib`? Any reason not to use that one?

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+    generated_deps_name = "_generated_{}_deps".format(name)
+
+    nixpkgs_package(
+        name = generated_deps_name,
+        nix_file_content = """
+{ nix_file }:
+let
+  nixpkgs = import <nixpkgs> {};
+  pythonExpr = import nix_file;
+  inherit (pythonExpr) python pkgs;
```
aherrmann, line 330: Perhaps better to leave those qualified to prevent confusion with `python` and `pkgs` provided by `nixpkgs`?

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+    value = drv // { _pythonModules = filterPythonModules drv.propagatedBuildInputs; };
+  });
+  startSet = toClosureFormat pkgs;
+  closure = builtins.genericClosure {
+    inherit startSet;
+    operator = item: toClosureFormat (filterPythonModules item.value.propagatedBuildInputs);
+  };
+
+  # Using the information generated above, map the package information into
+  # a list, described in the Output description at the top of this file.
```
aherrmann, line 355: Which "Output description" does this refer to?

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
 
     if register:
         native.register_toolchains("@{}//:toolchain".format(name))
+
+
+def _nixpkgs_python_repository_impl(repository_ctx):
```
aherrmann, line 205: Hmm, maybe there's a better name for this. Perhaps `_nixpkgs_python_poetry_packages?

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+    # 2. read generated json
+    python_modules = repository_ctx.read(repository_ctx.path(repository_ctx.attr.json_deps))
+
+    content = 'load("//:python_module.bzl", "python_module");'
```
aherrmann, line 209: Again related to the Q/A discussion. Since this file is the same every time it'd be better to include it in `rules_nixpkgs_python` as a source and just load it from there. 
```suggestion
    content = 'load("@rules_nixpkgs_python//:python_module.bzl", "python_module");'
```

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+    files = glob(["{link}/**"], exclude=["{link}/**/* *"]),
+    deps = {deps},
+    visibility = ["//visibility:public"],
+)
+        """.format(name=pkg_name, link=pkg_link_path, deps=deps)
+
+    repository_ctx.file("BUILD.bazel", content)
+
+    # 3. generate BUILD.bazel file,
+    # ... _and_ the symlinks
```
aherrmann, line 232: Should this comment go up a bit?

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+    # 4. Generate dummy WORKSPACE
+    # repository_ctx.file("WORKSPACE", "")
```
aherrmann, line 234:
```suggestion
```
Looks like a left-over. Bazel will generate an empty `WORKSPACE` by default.

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+
+    repository_ctx.file("python_module.bzl", """
+def _python_module_impl(ctx):
+    import_depsets = []
+    store = ctx.file.store_path
+    runfiles = ctx.runfiles(files = [store])
+
+    for dep in ctx.attr.deps:
+        runfiles = runfiles.merge(dep[DefaultInfo].data_runfiles)
+        import_depsets.append(dep[PyInfo].imports)
```
aherrmann, line 244: [`data_runfiles` is deprecated and its use discouraged](https://bazel.build/extending/rules#deprecated_features), use `default_runfiles` and explicitly include required outputs instead.

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+
+    # HACK(danny): for some unforunate reason, short_path returns ../ when operating in external
+    # repositories. I don't know why. It breaks rules_python's assumptions though.
+    fixed_path = store.short_path[3:]
+    import_path = "/".join([ctx.workspace_name, store.short_path])
```
aherrmann, line 249: A bit of context for this issue can be found in https://github.com/bazelbuild/bazel-skylib/issues/303

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+
+    return [
+        DefaultInfo(
+            files = depset(ctx.files.files),
+            default_runfiles = ctx.runfiles(ctx.files.files, collect_default = True),
```
aherrmann, line 254: Likewise, [use of `collect_data` and `collect_default` is discouraged](https://bazel.build/extending/rules#deprecated_features).

toolchains/python/python.bzl
```diff
@@ -200,3 +200,184 @@ def nixpkgs_python_configure(
...
+    )
+
+    # 5. generate //:requirements.bzl for later import.
+    repository_ctx.file("requirements.bzl", """
+def requirement(package_name):
+    return "@{}//:{{}}".format(package_name)
+""".format(repository_ctx.name),
+    )
+
+    # TODO: make it lazy in the packages themselves ?
```
aherrmann, line 292: I think that would require an extra indirection. I.e. generate a `.bzl` file that defines a macro that invokes `nixpkgs_package` for each package. I'm happy to leave this open as a future improvement.

aherrmann (commented):
Thank you for pushing this forward! This is a really cool feature, and I don't think much is missing to have this ready to merge. I've added a couple comments.

testing/python-poetry/MODULE.bazel
```diff
@@ -0,0 +1,52 @@
+module(name = "rules_nixpkgs_python_testing")
```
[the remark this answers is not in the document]
layus: I do not really understand how tests are organized. That was my attempt not to break everything ;-). So I will merge them.

toolchains/python/python.bzl
layus: The file is not generated, no. But this flexibility adds complexity to the expression.

https://github.com/tweag/rules_nixpkgs/blob/3748711f010bd191b0867344d615f9dc98b6eaf9/toolchains/cc/cc.nix#L48-L55

Given that it is unlikely to have more than one such package set, I am happy to require users to write their expression to a file if they really want an expression.

For now I would like to keep it as-is.

toolchains/python/python.bzl
layus: nixpkgs.lib has one, yes. Not builtin though. Initially I wanted this to be nixpkgs free. Seems I needed writeTextFile in the end, but I would rather get rid of it. Importing all of nixpkgs just for that helper seems a bit overkill.

layus: And this one uses the fact that values can be converted to dictionary keys to speed up the process. nixpkgs' version is O(n^2).

toolchains/python/python.bzl
layus: Leftover from [redacted] :-D.

toolchains/python/python.bzl
layus: It is a generic implementation. The fact that poetry/poetry2nix have been used to generate the set is not important for the rule. It's generic on purpose. I think the updated description will be enough to clarify this.

testing/python-poetry/MODULE.bazel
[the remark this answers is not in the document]
aherrmann: For context, it's explained in https://github.com/tweag/rules_nixpkgs/pull/319 and the [corresponding README](https://github.com/tweag/rules_nixpkgs/blob/master/testing/README.md).

toolchains/python/python.bzl
aherrmann: Fair enough. Maybe `_nixpkgs_python_packages` then? With `nixpkgs_python_repository` it could also refer to a Python interpreter repository. Just trying to avoid the ambiguity.

toolchains/python/python.bzl
layus: Like an export_files, or a bzl_library ?

toolchains/python/python.bzl
layus: Now I remember, the core problem here is remote execution, as the symlink gets flattened to it's exposed content. I may need some peer assistance to get all of this right.

layus: I think I got this mostly right in the last commit.

layus:
Stuck on some disk space limitation. Probably /run/usr/${id} AFAIR, because nix uses that as $TMP and mach-nix fills it by unpacking it's database of all the python packages it knows.

toolchains/python/python.bzl
aherrmann: Shouldn't be needed for the `load`. At the moment `bzl_library` is only needed for Stardoc and this generated file will not have Stardoc generated for it. `export_files` is only needed if you want to make the file a source for a rule in another BUILD file.

aherrmann:
This provides a solution to https://github.com/tweag/rules_nixpkgs/issues/346 for Python.

layus:
Done, I think it is as ready as can be.

README.md
```diff
@@ -1836,6 +1837,153 @@ default is <code>True</code>
...
+While depending on such a label directly will work, the layout may change
+in the future. To be on the safe side, define and import your own
+`requirement` function if you need to play with these labels.
+
+:warning: Just as with rules_python, nothing is done to enforce consistency
+between the version of python used to generate this repository and the one
+configured in your toolchain, even if you use nixpkgs_python_toolchain. You
+should ensure they both use the same python from the same nixpkgs version.
+While nixpkgs_python_repository could also register the toolchain used to
+generate the package set, this is not implemented yet.
```
aherrmann, line 1884: And inconsistencies would still be possible if another toolchain is registered next to it. So, it's unclear if we really want to do this.

testing/python/tests/import_packages_test.py
```diff
@@ -0,0 +1,11 @@
...
+print("python path = ", end="")
+pprint.pprint(sys.path)
+
+import flask
+print("flask version = ", flask.__version__)
+
+# Also test that numpy is available.
+# nixpkgs_python_repository should provide the dependencies closure.
+import click
+print("click version = ", click.__version__)
```
aherrmann, line 11: Is the comment out of date, or how is click related to numpy?

toolchains/python/package_set_to_json.nix
```diff
@@ -0,0 +1,48 @@
+{
+  # A file to be later `import`ed, and providing two attributes: 
+  # - `python`: used to obtain the valie of `sitePackages`
```
aherrmann, line 3:
```suggestion
  # - `python`: used to obtain the value of `sitePackages`
```

README.md
aherrmann:
```suggestion
```
Maybe better to remove this here to not set up false expectations. Especially in lite of the upcoming bzlmod extensions, there may be better ways to deal with this in the bzlmod world.

aherrmann (approved):
Thank you @layus , this is great! 
I've added a few more comments, but all of them are small comments.
Please address the comments and then feel free to add the merge-queue label.

layus:
Concerning mach-nix, it is way too difficult to make it work on recent nixpkgs, the project is currently unmaintained, and last update it received references a nixpkgs from one year ago. I remembered just in time that you can craft python packages sets manually in nixpkgs.
Once dream2nix support for python improves, we could also add it to the list. DavHau moved it's focus from mach-nix to dream2nix, so I am pretty confident dream2nix will make steady progress on python. I will never thank that person enough :sparkles:

layus:
@aherrmann Thanks for your thorough and fast reviews, working on this PR has been a real pleasure !

aherrmann:
> Once dream2nix support for python improves, we could also add it to the list. DavHau moved it's focus from mach-nix to dream2nix, so I am pretty confident dream2nix will make steady progress on python. I will never thank that person enough sparkles

That sounds very exciting! Yes, perhaps there is even a way to use dream2nix more broadly for rules_nixpkgs integration of different languages.

> Thanks for your thorough and fast reviews, working on this PR has been a real pleasure !

Likewise, thank you for the great work!

PR #350, opened 2023-03-14
Fix arguments in cc_configure

dmadisetti:
--arg must be argstr to allow for arguments such as 'c++'
More over, allow for ccLang to be passed any path. Only prescribing attribute path is undocumented, and unexpected

### DIFF
--- a/toolchains/cc/README.md
+++ b/toolchains/cc/README.md
@@ -10,7 +10,7 @@ One may wish to use a C++ toolchain to compile certain libraries written in
 non-C++ languages. For instance, Clang/LLVM can be used to compile CUDA or HIP
 code targeting GPUs. This can be achieved by:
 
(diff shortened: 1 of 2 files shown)


dmadisetti:
@evertedsphere, since I think you were here last in #298

layus:
Hi, thanks for your PR :+1: 
Technically, an --arg will always allow you more flexibility than an --argstr. You can pass `--arg '"c++"' ` to pass a value that happens to be a nix string, and still allow for other kind of expressions.
I just launched CI for now, and will have a closer look later.

benradf (approved):
@layus is correct, but in this case I think `ccLang` should always be a string rather than a general nix expression, so it makes sense to use `--argstr`. Thanks for the PR!

PR #352, opened 2023-03-15
ci(Mergify): configuration update

layus:
This change has been made by @layus from Mergify config editor.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,5 +1,6 @@
 queue_rules:
   - name: default
+    merge_method: merge
     conditions:
       - check-success~=Build & Test - Nixpkgs - .*
       - check-success~=Build & Test - Examples \(.*\)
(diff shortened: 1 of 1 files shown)


layus:
Trying to fix the issue seen in https://github.com/tweag/rules_nixpkgs/pull/350 where mergify did not kick in (https://github.com/tweag/rules_nixpkgs/pull/350/checks?check_run_id=12016141108) due to the fact that mergify now fails instead of using a random org user to perform the merge, as announced in https://changelog.mergify.com/changelog/rebasing-pull-requests-with-a-random-user-is-deprecated. At least that is what I learned in https://github.com/Mergifyio/mergify/issues/5075. Another option would be to use a more advanced merge method, as our paid plan allows, but we need a bot account for that. Is there an official Tweag bot for that ?

benradf (approved):
Thanks for fixing this!

PR #369, opened 2023-04-12
Implement remaining nix_repo tags

aherrmann:
nix_repo_tags


testing/core/tests/BUILD.bazel
```diff
@@ -57,9 +44,9 @@ expand_location_unit_test_suite()
         srcs = ["test_output.sh"],
         args = [
             "2",
-            "$(locations @output-filegroup-test//:include)",
+            "$(locations {})".format(nix_pkg("rules_nixpkgs_core_testing", "output-filegroup-test", "//:include")),
         ],
-        data = ["@output-filegroup-test//:include"],
+        data = [nix_pkg("rules_nixpkgs_core_testing", "output-filegroup-test", "//:include")],
```
benradf, line 49: I was a bit confused why the `nix_pkg` module extension was being used like this, but now I see `nix_pkg` here is a helper macro from `defs.bzl`.

core/extensions/package.bzl
```diff
@@ -38,45 +38,115 @@ def _name_from_attr(attr):
...
+    )
+
+def _local_file_pkg(key, local_file):
+    kwargs = _handle_common_attrs(local_file)
+    kwargs.update(_handle_repo_attrs(key, local_file))
+    kwargs.update(_handle_build_attrs(local_file))
+    kwargs.update(_handle_file_attrs(local_file))
+    kwargs.update(_handle_opts_attrs(local_file))
+
+    # Inidicate that nixpkgs_package is called from a module extension to
```
benradf, line 134: `s/Inidicate/Indicate/`

benradf (approved):
Looks good :+1:

core/extensions/package.bzl
aherrmann:
```suggestion
    # Indicate that nixpkgs_package is called from a module extension to
```

testing/core/tests/BUILD.bazel
aherrmann: Probably something to make clear in the docs. Or, if we end up getting `local_use_repo` the problem would go away.

Issue #348, opened 2023-03-13
Future direction of rules_nixpkgs regarding nix flakes


benradf:
**Is your feature request related to a problem? Please describe.**
With the Nix ecosystem moving towards [flakes](https://nixos.wiki/wiki/Flakes) we should start thinking about what this means for rules_nixpkgs. Flakes are still experimental in Nix and will need to stabilise before we can fully commit to an approach, but starting the discussion now and exploring ideas will help us prepare.

Some things to think about:
- How can flakes be integrated into rules_nixpkgs?
- How can they benefit users?
- What API improvements can they enable?
- etc

Discussions on the subject:
-  [Flakes and support for Nix evaluation cache](https://github.com/tweag/rules_nixpkgs/discussions/341)
- [Flakes first-class support](https://github.com/tweag/rules_nixpkgs/discussions/342)

rickvanprim:
The current approach I'm attempting to use is having a Flake at the `WORKSPACE` level that defines both the `devShell` and a `bazel` set under `packages."${system}"` containing any derivations I want to pull into Bazel.  From there, I quickly wrote `nix_flake_package()` (using `nixpkgs_package()` as a starting point) that uses `nix build` to get the output path for a given derivation.
Example:
```bazel
nix_flake_package(
    name = "nix_rust",
    nix_flake_file = "//:flake.nix",
    nix_flake_lock_file = "//:flake.lock",
    package = "bazel.rust",
)
```
This expand to `nix build <nix_flake_file.dirname>#bazel.rust --print-out-paths --out-link bazel-support/nix-out-link <nixopts>` and then uses the same logic as `nixpkgs_package()`.  One of the main upsides I see is not needing the `repository`/`repositories` parameters of `nixpkgs_package()` as the Flake evaluation takes care of all of that and Bazel doesn't have to separately fetch them.

Rough docs:
```
Make the content of a local Nix Flake package available in the Bazel workspace.

Args:
  name: A unique name for this repository.
  nix_flake_file: Label to `flake.nix` that will be evaluated.
  nix_flake_lock_file: Label to `flake.lock` that corresponds to `nix_flake_file`.
  nix_flake_file_deps: Additional dependencies of `nix_flake_file` if any.
  package: Nix Flake package to make available.  The default package will be used if not specified.
  ... <build_file docs+behavior copied from `nixpkgs_package()`
```

If this approach/API seems reasonable, I'd be happy to put up a PR 🙂

[...]

PR #380, opened 2023-05-05
feature: nixpkgs_flake_package

rickvanprim:
Introduces a new repository rule `nixpkgs_flake_package` that invokes `nix build <label>#<package>` and makes the output available as a Bazel repository.  This allows all the Nix version pinning to happen as part of the Nix Flake functionality with `flake.lock` and simplifies the Bazel side implementation by not having to fetch `nixpkgs` (or any other Nix repositories) outside of Nix and inject them in.

Contributes to #348.


core/nixpkgs.bzl
```diff
@@ -1,4 +1,4 @@
-"""<!-- Edit the docstring in `core/nixpkgs.bzl` and run `bazel run //docs:update-README.md` to change this repository's `README.md`. -->
+"""<!-- Edit the docstring in `core/nixpkgs.bzl` and run `bazel run @rules_nixpkgs_docs//:update-readme` to change this repository's `README.md`. -->
```
rickvanprim, line 1: Updated to match `//nixpkgs/nixpkgs.bzl`.

core/nixpkgs.bzl
```diff
@@ -453,6 +454,7 @@ def _nixpkgs_package_impl(repository_ctx):
     else:
         # No user supplied build file, we may create the default one.
         create_build_file_if_needed = True
+
```
rickvanprim, line 457: Auto-added by Buildifier.

core/nixpkgs.bzl
```diff
@@ -668,3 +670,224 @@ def nixpkgs_package(
...
+    elif repository_ctx.attr.build_file:
+        repository_ctx.symlink(repository_ctx.attr.build_file, "BUILD")
+    elif repository_ctx.attr.build_file_content:
+        repository_ctx.file("BUILD", content = repository_ctx.attr.build_file_content)
+    else:
+        # No user supplied build file, we may create the default one.
+        create_build_file_if_needed = True
+
+        # Workaround to bazelbuild/bazel#4533
+        repository_ctx.path("BUILD")
```
rickvanprim, line 717: This could potentially be pulled out of here and `nixpkgs_package()`.  I don't think there's a good way to deduplicate the docstring though?

core/nixpkgs.bzl
```diff
@@ -668,3 +670,224 @@ def nixpkgs_package(
...
+
+    nix_flake_file_deps = {}
+    for dep_lbl, dep_str in repository_ctx.attr.nix_flake_file_deps.items():
+        nix_flake_file_deps[dep_str] = cp(repository_ctx, dep_lbl)
```
rickvanprim, line 721: This could potentially be grouped with the `expand_location()` call below and pulled out to be shared with `nixpkgs_package()`.

core/nixpkgs.bzl
```diff
@@ -668,3 +670,224 @@ def nixpkgs_package(
...
+    for target in find_children(repository_ctx, output_path):
+        basename = target.rpartition("/")[-1]
+        repository_ctx.symlink(target, basename)
+
+    # Create a default BUILD file only if it does not exists and is not
+    # provided by `build_file` or `build_file_content`.
+    if create_build_file_if_needed:
+        p = repository_ctx.path("BUILD")
+        if not p.exists:
+            repository_ctx.template("BUILD", Label("@rules_nixpkgs_core//:BUILD.bazel.tpl"))
```
rickvanprim, line 796: Could potentially be pulled into the BUILD file function mentioned above.

core/nixpkgs.bzl
```diff
@@ -668,3 +670,224 @@ def nixpkgs_package(
...
+                                        """),
+    },
+)
+
+def nixpkgs_flake_package(
+        name,
+        nix_flake_file,
+        nix_flake_lock_file,
+        nix_flake_file_deps = [],
+        package = None,
```
rickvanprim, line 820: Would `attribute_path` be a better name than `package`?  That would be consistent with `nixpkgs_package()`.  I chose `package` here initially as it matches the name of one of the known Flake output fields.

rickvanprim:
Looking at `//tests`, it looks like there's only really one test right now.  Should I copy-paste `invalid_nixpkgs_package` (directory and shell script) to make a test for `nixpkgs_flake_package` or what's the expectation as far as adding tests? 🙂

core/nixpkgs.bzl
```diff
@@ -668,3 +670,224 @@ def nixpkgs_package(
...
+
+    Args:
+      name: A unique name for this repository.
+      nix_flake_file: Label to `flake.nix` that will be evaluated.
+      nix_flake_lock_file: Label to `flake.lock` that corresponds to `nix_flake_file`.
+      nix_flake_file_deps: Additional dependencies of `nix_flake_file` if any.
+      package: Nix Flake package to make available.  The default package will be used if not specified.
+      build_file: The file to use as the BUILD file for this repository.
+
+        Its contents are copied copied into the file `BUILD` in root of the nix output folder. The Label does not need to be named `BUILD`, but can be.
```
aherrmann, line 837:
```suggestion
        Its contents are copied into the file `BUILD` in root of the nix output folder. The Label does not need to be named `BUILD`, but can be.
```

core/nixpkgs.bzl
aherrmann: Factoring out the common parts between `nixpkgs_package` and `nixpkgs_flake_package` would be great. There is a fair bit of complexity around invoking Nix that was tricky to get right. I'd be wary of duplicating it and risk it diverging.

benradf:
The `workspace` CI jobs are failing because the README doesn't match the docstrings in the bzl files. We shouldn't change the README files directly - there are targets to regenerate them from the docstrings. This is minor though - we can fix it when the PR is otherwise ready merge.

The `bzlmod` CI jobs are failing with this error:
```
Error in fail: Invalid Nix repository, you must use the nix_repo extension and request a global repository or register a local repository: Unknown repository - no repository named 'flake-hello' available for module 'rules_nixpkgs_core_testing'.
```
You can reproduce this locally by running the following in the `testing/core` directory:
```
bazel run --config=bzlmod //tests:run-flake-hello
```
I started looking into how the bzlmod code path would need to change to fix this but didn't figure it out yet. I'm guessing `nixpkgs_flake_package` doesn't do some `bzlmod` stuff that `nixpkgs_package` does. Any suggestions @aherrmann?

testing/core/tests/BUILD.bazel
```diff
@@ -31,6 +31,8 @@ expand_location_unit_test_suite()
         "expr-attribute-test",
         "relative-imports",
         "extra-args-test",
+        "flake-hello",
+        "flake-hello-with-build-file",
```
aherrmann, line 35: In reply to https://github.com/tweag/rules_nixpkgs/pull/380#issuecomment-1543902257.
The issue is that these tests were added here. This list comprehension is only valid for tests of packages that are imported using a bzlmod Bazel module extension. However, no such extension exists for the flake part, yet.
You can fix this by adding another instance of the list comprehension like this (unverified code):
```
[
    # All of these tests use the "hello" binary to see
    # whether different invocations of `nixpkgs_flake_package`
    # produce a valid bazel repository.
    sh_test(
        name = "run-{0}".format(test),
        timeout = "short",
        srcs = ["test_bin.sh"],
        args = ["$(location @{0}//:bin)".format(test)],
        data = ["@{0}//:bin".format(test)],
    )
    for test in [
        "flake-hello",
        "flake-hello-with-build-file",
```
(If you go through the Git history of this file, you'll find instances of such code from when not all tests had been converted to use the module extension, yet)

You'll also need to add `use_repo` entries to `MODULE.bazel` to import these two repositories from the `non_module_deps` extension.

rickvanprim: Is that just a temporary workaround or the proper fix?  It seems like my work here is only half done, as I presumably need to write `@rules_nixpkgs_core//extensions:flake_package.bzl` (or whatever bikeshed)?  This is based on my observation that `nix_pkg = use_extension("@rules_nixpkgs_core//extensions:package.bzl", "nix_pkg")` seems to be a bzlmod wrapper/validator around `nixpkgs_package()`.

rickvanprim: bzlmod test fixes pushed.  Not sure if they work as the CI jobs don't seem to run automatically?

benradf: Yeah, we have to run them manually for contributors from outside the organization. Looks like CI is passing now.

benradf (approved):
Good to see `_nixpkgs_package_impl` being broken up a little - it was getting pretty big for a single macro. Thanks for this contribution @rickvanprim!

PR #385, opened 2023-05-11
go 1.20 support

ylecornec:
This PR adds support for go 1.20 to `nixpkgs_go_configure`.

Similarly to what was done in rules_go [here](https://github.com/bazelbuild/rules_go/pull/3385/files#diff-6b36cd5e735033755d6df88acbc31780b4792fb54cec66a592966aba0f275ea1):
- It allows the glob that selects the standard library folder to be empty (because this folder does not exist anymore in version 1.20).
- It passes the version of `go` to the `go_sdk` rule.

In order to recover the version from the nix expression defining the toolchain, we use a `go.nix` wrapper file (similarly to what is done for the cc toolchain) which will add a `go_version.bzl` file to the toolchain.

A breaking change is that derivations passed to `nixpkgs_go_configure` via `nix-file`, `nix-file-content` or `attribute_path` must now contain a `version` attribute.
However I would expect most existing occurrences to use the `attribute_path` of a derivation already containing a `version` attribute (this is the case for the default `go` value).

### DIFF
--- a/toolchains/go/go.nix
+++ b/toolchains/go/go.nix
@@ -1,0 +1,38 @@
+# Exactly one of goAttrPath or goExpr must be set.
+# If a custom `goExpr` derivation is provided by the user, we add the `go_version.bzl` file to it.
+# Otherwise we create a toolchain based on the `goAttrPath` attribute path from <nixpkgs>
+
+let
+  pkgs = import <nixpkgs> {
+    config = { };
+    overlays = [ ];
+  };
+in args@{ goAttrPath ? null, goExpr ? null }:
+let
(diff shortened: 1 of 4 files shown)


toolchains/go/go.nix
```diff
@@ -0,0 +1,38 @@
+# Exactly one of goAttrPath or goExpr must be set.
+# If a custum `goExpr` derivation is provided by the user, we add the `go_version.bzl` file to it.
```
benradf, line 2: _custom_

benradf (approved):
Looks good!

PR #393, opened 2023-06-09
Update rust toolchain to use rules_rust 0.22.0

contagnas:
rules_rust [0.22.0](https://github.com/bazelbuild/rules_rust/releases/tag/0.22.0) removes rust_toolchain.os in https://github.com/bazelbuild/rules_rust/pull/1960

note this change isn't compatible with earlier versions of rules_rust, because the os field was required.


toolchains/rust/MODULE.bazel
```diff
@@ -1,6 +1,6 @@
 module(
     name = "rules_nixpkgs_rust",
-    version = "0.9.0",
+    version = "0.10.0",
 )
```
aherrmann, line 4: Let's defer the version bump here to when we actually cut a new release.

aherrmann (changes requested):
Thank you!
Small comment about the version bump.

benradf:
Thanks for the PR @contagnas!

It looks like [CI is failing](https://app.buildbuddy.io/invocation/6daa1a2a-f40f-44ac-a40c-eab0a9bc9992) with this error:
```
ERROR: ... BUILD:66:15: @nixpkgs_config_rust//:rust_nix_impl: missing value for mandatory attribute 'os' in 'rust_toolchain' rule
```
Which is odd, if `os` has been removed in rules_rust 0.22.0. I haven't had a chance to dig deeper yet.

registry/modules/rules_rust/metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "homepage": "https://bazelbuild.github.io/rules_rust/",
   "maintainers": [],
-  "versions": ["0.17.0"],
+  "versions": ["0.22.0"],
```
contagnas, line 4: I'm not familiar with bzlmod yet, so it's not clear to me if this registry should have multiple version of dependencies, or just the current one?

registry/modules/rules_rust/0.22.0/source.json
```diff
@@ -1,5 +1,5 @@
 {
-  "url": "https://github.com/bazelbuild/rules_rust/releases/download/0.17.0/rules_rust-v0.17.0.tar.gz",
+  "url": "https://github.com/bazelbuild/rules_rust/releases/download/0.22.0/rules_rust-v0.22.0.tar.gz",
   "integrity": "sha384-v3eVoRi4rdkELxzV3egjTJkc8M3UH1YmbdZ2m6Y45gpjBew4a/JU5A4TFNO8Bq3z",
```
contagnas, line 3: I expected this sha to need an update, but I can't find anything that uses it

registry/modules/rules_rust/metadata.json
aherrmann: This registry is only used for testing (see [readme](https://github.com/tweag/rules_nixpkgs/tree/0c1f8f5470c7f292b7620762e224f53d837929d3/registry#readme)), so a single version is enough.

registry/modules/rules_rust/0.22.0/source.json
aherrmann: bazel checks against this sha after downloading the dependency. I think it will also cache based on that sha, so if it's not updated you may get an old artifact.

benradf:
A different CI error now :thinking:
```
   ERROR: /home/runner/work/rules_nixpkgs/rules_nixpkgs/examples/toolchains/rust/WORKSPACE:112:18: fetching crates_repository rule //external:crate_index: Traceback (most recent call last):
  	File "/home/runner/.cache/bazel/_bazel_runner/962535b5b03441543344ad7194339468/external/rules_rust/crate_universe/private/crates_repository.bzl", line 45, column 28, in _crates_repository_impl
  		repin = determine_repin(
  	File "/home/runner/.cache/bazel/_bazel_runner/962535b5b03441543344ad7194339468/external/rules_rust/crate_universe/private/generate_utils.bzl", line 382, column 13, in determine_repin
  		fail(("\n".join([
  Error in fail: Digests do not match: Digest("59ace66e05f26b5ab71bdc99a930d8b1bafdb232f70f60ad13c15599bc4a49a4") != Digest("a287c27b9422cf61b5995ee3c56060decae072616e470f667c74d3bd1650e80f")
  
  The current `lockfile` is out of date for 'crate_index'. Please re-run bazel using `CARGO_BAZEL_REPIN=true` if this is expected and the lockfile should be updated.
```

benradf:
The expected hash that gets printed when there's a mismatch is in hexadecimal, but the hash in `source.json` should be base64. So I had to do a conversion:
```
$ (xxd -r -p | base64) <<<f47303cd12b8b8ebfdce0e2ac93f2b72d1d484d7ef9bd6f3b4f307f523e331ebbf3496caed2688f4d5345918344be9f2
9HMDzRK4uOv9zg4qyT8rctHUhNfvm9bztPMH9SPjMeu/NJbK7SaI9NU0WRg0S+ny
```

benradf:
Fixed the failure in `examples/toolchains/rust` by updating the lockfile:
```
CARGO_BAZEL_REPIN=true bazel run --config=nix :hello
```

Issue #347, opened 2023-03-13
Better API for cross-compilation support


benradf:
**Is your feature request related to a problem? Please describe.**
In principle cross-compilation is already supported thanks to the nixpkgs `system` and `targetSystem` parameters. However this requires the user to write their own nix expressions and does not integrate nicely with how Bazel handles cross-compilation. Hence the need for a better API around this. 

**Describe the solution you'd like**
We can generate toolchain instances for a set of exec and target platform pairs and apply appropriate Bazel constraints automatically. This needs to be configurable. The [nodejs toolchain](https://github.com/tweag/rules_nixpkgs/blob/master/toolchains/nodejs/nodejs.bzl) is an example where this is somewhat supported.

For packages, we want to generate instances for a set of platforms. Build tool packages may incur separate exec and target constraints (i.e. a package with a binary that runs during the build and a shared library for the target platform). There should be an alias target that uses `select` to pick the appropriate package instance.

layus:
Just a quick note here: Aliases are cumbersome as you can only alias targets, not full-blown external repositories. So you need to know all the bazel targets declared in a nixpkgs_package to alias them.

aherrmann:
@layus Yes, that's true - and annoying. Unfortunately, I'm not sure what a better approach may look like. Short of defining toolchains.

DolceTriade:
I'm interesting in tackling this. I put together a hacky proof of concept of a pretty involved repository set up that works off of https://github.com/tweag/rules_nixpkgs/pull/240 where I can cross compile for linux x86_64 on my M1 macbook, and then run the resulting binaries in Docker using `--run_under`.

There are only minor changes required to rules_nixpkgs (namely, that the toolchain CPU for C++ is hard coded to the system CPU).

I ended up with something like this in my `WORKSPACE.bazel`
```
nixpkgs_cc_configure(
    name = "nixpkgs_crossosx_cc",
    cross_cpu = "k8",
    exec_constraints = ["@platforms//os:osx", "@platforms//cpu:arm64"],
    nix_file = "//bazel/nix/cc:osxcross_cc.nix",
    nixopts = [
        "--arg",
        "ccPkgs",
        "import <nixpkgs> { crossSystem = \"x86_64-linux\";}",
        "--show-trace",
    ],
    repository = "@nixpkgs",
    target_constraints = [
        "@platforms//cpu:x86_64",
        "@platforms//os:linux",
[...]

PR #401, opened 2023-08-13
cc: Add cross_cpu arg to nixpkgs_cc_toolchain_config

DolceTriade:
Currently, we hardcode the CPU architecture of the toolchain to the host toolchain. However, this breaks cross compilation as we will probably want to set the CPU of the toolchain to the target architecture. The `cross_cpu` flag can be used to specify the target architecture. If the `cross_cpu` flag is not set, we will auto detect the host CPU architecture and use that for the toolchain.

This is part of an effort to address #347


DolceTriade:
Ah, thanks. Updated.

DolceTriade:
Hm, I'm not certain why my changes aren't showing up here: https://github.com/DolceTriade/rules_nixpkgs/commit/17457f67509b69f58d774c9ca47a684559d75c7e

maybe I should close and re-open...

benradf:
Not sure why CI isn't being triggered now (I don't even see an option to start it manually). Maybe the force push confused it?

I'll try closing the pull request and opening it again.

PR #404, opened 2023-08-20
Add support for nix store remote copying

z8v:
This is a requirement in order to make Bazel remote execution work. Without it it's not guaranteed that any necessary nix paths will be available during the build.

Any local derivation will be copied over to a specified remote nix server which will also as a Bazel executor. This way any nix paths required during the build, will be available.

The user will have to provide the `BAZEL_NIX_REMOTE` environment variable. This should be the name of any entry on the SSH_CONFIG file where all the authentication details are provided.

e.g

```bash
$ export BAZEL_NIX_REMOTE=nix-server

$ cat ~/.ssh/config
Host nix-server
  Hostname 1.2.3.4
  IdentityFile ~/.ssh/nix-server
  Port 2222
  User nix-user
```

This was done in order to simplify the processes of authentication and keep the number of configuration options to a minimum.

### DIFF
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -429,7 +429,7 @@ def _nixpkgs_build_and_symlink(repository_ctx, nix_build_cmd, expr_args, build_f
         repository_ctx.report_progress("Creating remote store root")
         exec_result = execute_or_fail(
             repository_ctx,
-            [ssh_path] + [nix_host, "nix-store --add-root /nix/var/nix/gcroots/per-user/nix/rules_nixpkgs_{root} -r {path}".format(root = output_path.split('/')[-1], path = output_path) ],
+            [ssh_path] + [nix_host, "nix-store --add-root ~/rules_nixpkgs_gcroots/{root} -r {path}".format(root = output_path.split('/')[-1], path = output_path) ],
             failure_message = "Cannot create remote store root for Nix attribute '{}'.".format(
                 repository_ctx.attr.name,
             ),
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -431,7 +431,6 @@ def _nixpkgs_build_and_symlink(repository_ctx, nix_path, expr_args, build_file_c
             timeout = 10000,
         )
 
-        nix_path = repository_ctx.which("nix")
         repository_ctx.report_progress("Downloading Nix derivation")
         exec_result = repository_ctx.execute(
             [nix_path, "copy", "--from", nix_store, output_path],
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -396,21 +396,59 @@ def _nixpkgs_build_file_content(repository_ctx):
     else:
         return None
 
-def _nixpkgs_build_and_symlink(repository_ctx, nix_build, build_file_content):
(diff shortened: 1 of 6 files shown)


core/nixpkgs.bzl
```diff
@@ -396,22 +396,60 @@ def _nixpkgs_build_file_content(repository_ctx):
...
+             quiet = repository_ctx.attr.quiet,
+             timeout = timeout,
+        )
+        output_path = exec_result.stdout.splitlines()[-1]
+
+        ssh_path = repository_ctx.which("ssh")
+        repository_ctx.report_progress("Creating remote store root")
+        exec_result = execute_or_fail(
+            repository_ctx,
+            [ssh_path] + [nix_host, "nix-store --add-root /nix/var/nix/gcroots/per-user/nix/rules_nixpkgs_{root} -r {path}".format(root = output_path.split('/')[-1], path = output_path) ],
```
benradf, line 426: The gcroot path should probably be configurable, since the ssh user might not have permission to write a symlink here. Another approach might be to rely on convention and put the root in the home folder of the ssh user, e.g.
```
nix-store --add-root ~/rules_nixpkgs_gcroots/{root} -r {path}
```

z8v: Updated with the suggestion. Although the nix user will be able to write on `/nix/store` so it should be able also to write on that path also.

z8v:
@benradf I believe it would make sense to include in this PR a docker/vagrant setup to test the remote build on the CI?

core/nixpkgs.bzl
benradf: Ah yeah, if you ssh in as an actual nix user that might be the case. Still, there are various ways Nix can be configured (e.g. `local` store vs `daemon` store), so it's probably best to assume as little as possible about what can be done on the remote.

benradf:
That sounds like a great idea actually, though I'm not entirely sure how to go about it.

What were you thinking? Build a docker image as part of CI, run it, and test that it can be used as a remote executor?

z8v:
> That sounds like a great idea actually, though I'm not entirely sure how to go about it.
> 
> What were you thinking? Build a docker image as part of CI, run it, and test that it can be used as a remote executor?

I was think to have a container that will play the role for the nix server. We could run this on the CI so we can test the code path that is enabled when we specify a remote nix server.

benradf (approved):
Maybe add something like `${{ matrix.withNixRemote && '- NixRemote' || '' }}` to the end of the [job name](https://github.com/tweag/rules_nixpkgs/pull/404/files#diff-fde0e5d64aae13964fdda6d47af304cf1a7015cbc17e440ac4a5e662ee1d875eR20) in `workflow.yaml` to distinguish jobs on the `withNixRemote` dimension.

Otherwise, looks good :+1:

z8v:
> Maybe add something like `${{ matrix.withNixRemote && '- NixRemote' || '' }}` to the end of the [job name](https://github.com/tweag/rules_nixpkgs/pull/404/files#diff-fde0e5d64aae13964fdda6d47af304cf1a7015cbc17e440ac4a5e662ee1d875eR20) in `workflow.yaml` to distinguish jobs on the `withNixRemote` dimension.
> 
> Otherwise, looks good 👍

Updated the title. We'll need to adjust the required checks on Github with the new name.

benradf (approved):

PR #413, opened 2023-09-06
Fix `darwin` flag when cross-compiling from Linux to MacOS

benradf:
In response to [this comment](https://github.com/tweag/rules_nixpkgs/pull/402#discussion_r1316920564) from @jcpetruzza.

### DIFF
--- a/examples/toolchains/cc_cross_osx_to_linux_amd64/README.md
+++ b/examples/toolchains/cc_cross_osx_to_linux_amd64/README.md
@@ -1,13 +1,18 @@
-C++ With Dependencies Toolchain Example
-=======================================
+Linux C++ In Docker, Cross Compiled on MacOS
+============================================
 
-This is an example C++ project with dependencies that uses `rules_cc`.
-
-This example uses the Nix package manager to provide C++ dependencies, and as such only works with Nix installed. Demonstrating other methods of providing C++ dependencies is out of scope of this example.
+Builds a Docker image containing a C++ program cross compiled for Linux, on MacOS.
 
 # Usage
 
-To run the example with Nix, issue the following command:
+To build the image with Nix, issue the following command:
 ```
-nix-shell --command 'bazel run --config=cross :hello'
+nix-shell --command 'bazel build --config=cross :hello_image_tarball'
 ```
+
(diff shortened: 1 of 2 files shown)


toolchains/cc/cc.bzl
```diff
@@ -122,8 +122,8 @@ def _parse_cc_toolchain_info(content, filename):
 
 def _nixpkgs_cc_toolchain_config_impl(repository_ctx):
     host_cpu = get_cpu_value(repository_ctx)
-    cross_cpu = repository_ctx.attr.cross_cpu or host_cpu
-    darwin = (host_cpu == "darwin" or host_cpu == "darwin_arm64") and cross_cpu == host_cpu
+    cpu_value = repository_ctx.attr.cross_cpu or host_cpu
+    darwin = cpu_value == "darwin" or cpu_value == "darwin_arm64"
```
DolceTriade, line 126: Yes, I think this is correct and will ensure the correct version of _cc_wrapper is selected.

DolceTriade (approved):

examples/toolchains/cc_cross_osx_to_linux_amd64/README.md
```diff
@@ -1,13 +1,18 @@
...
 # Usage
 
-To run the example with Nix, issue the following command:
+To build the image with Nix, issue the following command:
+```
+nix-shell --command 'bazel build --config=cross :hello_image_tarball'
 ```
-nix-shell --command 'bazel run --config=cross :hello'
+
+You can then do load the image into Docker with:
```
aherrmann, line 13: It's also possible to just load the image with `bazel run`, [see here](https://github.com/bazelbuild/rules_docker/tree/master#using-with-docker-locally).

aherrmann (approved):
Thank you!

examples/toolchains/cc_cross_osx_to_linux_amd64/README.md
benradf: I was testing on our `build02` machine, which doesn't have Docker installed, but I think it's fair to assume Docker is there for the README. Will update.

PR #364, opened 2023-04-06
Implement remaining nix_repo tags

aherrmann:
Part of #183.

Implement the remaining `http`, `file`, and `expr` tags for `nix_repo`.
Ensure each of them is tested in `testing/core`.

Issue #183, opened 2022-01-28
Create module extensions


aherrmann:
Related to #181

**Is your feature request related to a problem? Please describe.**

The new [bzlmod][bzlmod] dependency management system introduces a new concept called [module extensions][module-extensions].
`rules_nixpkgs` currently only exposes repository rules. To fully "modularize" `rules_nixpkgs` we need to expose the same functionality that is currently exposed as repository rules in the form of module extensions.

**Describe the solution you'd like**

Bazel's repository rules are evaluated individually. I.e. each `nixpkgs_package` is evaluated separately and creates its own external repository. Module extensions on the other hand are evaluated collectively. Each invocation across all `MODULE.bazel` files in a given project and its transitive dependencies generates a tag with some metadata. In the end the corresponding module extension is evaluated with all these tags as inputs and can then use this information and invoke external tools to generate repository rule invocations to import all required external dependencies.

A direct translation of the current repository rules into a module extension could look something like this on the use-site:
```
bazel_dep(name = "rules_nixpkgs_core", version = "1.0")
nixpkgs = use_extension("@rules_nixpkgs_core//:extensions.bzl", "nixpkgs")

nixpkgs.git_repository(name = "nixpkgs", revision = "12.11", sha256 = ...)
nixpkgs.package(name = "boost", attribute_path = "boost175", repository = "nixpkgs")

bazel_dep(name = "rules_nixpkgs_cc", version = "1.0")
nixpkgs_cc = use_extension("@rules_nixpkgs_cc//:extensions.bzl", "nixpkgs_cc")
nixpkgs_cc.cc_configure(name = "nixpkgs_config_cc", repository = "nixpkgs")
use_repo(nixpkgs_cc, "nixpkgs_config_cc")

module(
    name = "user_project",
    ...
    toolchains_to_register = ["@nixpkgs_config_cc_toolchains//:toolchain"],
)
```
Note, the above is only a sketch and may not be possible in exactly that form.

**Describe alternatives you've considered**

The sketch above is a direct translation of the features provided by `rules_nipxkgs` right now. However, due to the collective evaluation of module extensions with bzlmod, we could go further and also import transitive Nix dependencies into Bazel in a way that handles overlapping dependency graphs correctly. E.g. if a project imports two C libraries `libA` and `libB` from `nixpkgs` and both of these libraries depend on `libC`, then the module extension could generate imports of all three `libA`, `libB`, and `libC` and expose the dependency graph to Bazel. E.g.
```
[...]

PR #423, opened 2023-09-29
Simplify rules_nixpkgs_core module extensions

aherrmann:
Part of #183.

The previous implementation of the rules_nixpkgs_core module extensions (https://github.com/tweag/rules_nixpkgs/pull/344, https://github.com/tweag/rules_nixpkgs/pull/363, https://github.com/tweag/rules_nixpkgs/pull/369, https://github.com/tweag/rules_nixpkgs/pull/364) used a [hub-repository](https://github.com/tweag/rules_nixpkgs/pull/364) to manage Nix repository and package import scope.

Since then, Bazel (>= 6.3) has introduced [isolated module extension instances](https://github.com/bazelbuild/bazel/pull/18529) and [automatic generation of `use_repo` stanzas](https://docs.google.com/document/d/1dj8SN5L6nwhNOufNqjBhYkk5f-BJI_FPYWKxlB3GAmA/edit?disco=AAAArdGBwhc). Both these features solve the same problems that the hub-repository implementation addressed, see below.

This PR replaces the previous hub-repository based implementation by a simpler implementation based on these new Bazel features. A clear advantage is that this means that users can use plain Bazel labels to refer to imported repositories and packages instead of having to access them through hub-repository accessors. This should also be an advantage in terms of WORKSPACE to bzlmod migration effort.

How isolated extensions and automatically generated `use_repo` address rules_nixpkgs's requirements:
- Separating globally unified and locally specialized packages. New: Globally unified behavior is the default, and locally specialized behavior is achieved using an isoalted module extension instance.
- Exposing imports under human friendly names. New: Because isolated module extension instances have their own namespace, we no longer need to worry about name mangling ourselves. Instead, we can generate external repositories with human friendly names directly, either the canonical attribute name in case of globally unified packages, or a human defined name in case of isolated instances or overrides in root or rules_nixpkgs_core. Thanks to the automatically generated `use_repo` stanzas, it is very easy for a user to manage the imports.


aherrmann:
The [second CI failure](https://github.com/tweag/rules_nixpkgs/actions/runs/6390957120/job/17345262106?pr=423#step:6:846) seems to be caused by an incompatibility with more recent nixpkgs revisions and MacOS 11 on x86_64. I have not been able to track down the exact reason, but worked around the issue by falling back to `pkgs.coreutils`. I've opened https://github.com/tweag/rules_nixpkgs/issues/424 to track this.

flake.nix
```diff
@@ -1,6 +1,6 @@
 {
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-22.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
```
layus, line 3: Any specific reason for this change ?

testing/core/tests/nixpkgs_repositories.bzl
```diff
@@ -10,29 +8,25 @@ load(
...
-        file_nixpkgs = nix_repo("rules_nixpkgs_core_testing", "file_nixpkgs")
-        nixpkgs_content = nix_repo("rules_nixpkgs_core_testing", "nixpkgs_content")
-    else:
-        nixpkgs = "@nixpkgs"
+    nixpkgs = "@nixpkgs"
+    remote_nixpkgs = "@remote_nixpkgs"
+    http_nixpkgs = "@http_nixpkgs"
+    file_nixpkgs = "@file_nixpkgs"
+    nixpkgs_content = "@nixpkgs_content"
+    if not bzlmod:
```
layus, line 16: XXX: Are all these "global" names defined in the documentation ?

tests/BUILD.bazel
```diff
@@ -8,7 +8,7 @@ sh_test(
         "//nixpkgs:srcs",
         "//tests/invalid_nixpkgs_package:srcs",
         "@coreutils_static//:bin",
-        "@nix_2_7//:bin",
+        "@nix_2_10//:bin",
```
layus, line 11: Any reason for thins change, except general maintenance ?

layus (approved):
Interesting changeset. Overall not much changed in the interface, except for some simplifications, so this is a much welcome change.

I would love to see the implications for more complex use cases of rules_nixpkgs, where @nixpkgs is not a vanilla upstream nixpkgs, but I think there is kind of a global assumption that @nixpkgs is some version of upstream nixpkgs.

I am still annoyed that bzlmod (at least for rules_nixpkgs) relies on so many conventions and unchecked assumptions, but there is not much we can do about that.

Some comments I left are merely notes for me to check after the review, these are marked with XXX.

flake.nix
aherrmann: Yes, to get access to Bazel version 6.3.2, which is not yet available in NixOS-22.11.

tests/BUILD.bazel
aherrmann: Yes, version 2.7 (and 2.8 and 2.9) are no longer available in the more recent nixpkgs revision (unstable). So we need to bump the version.

PR #433, opened 2023-10-18
Release 0.10.0

aherrmann:
Creates release 0.10.0.

- Adds a minimal testing module to the core component for testing on the Bazel Central Registry. This has to be included in a sub-directory of the module, see commit message.
- Add configuration for the add to BCR GitHub app.
- Write the Changelog for release 0.10.0.
- Bump the version to 0.10.0.
- Add a GitHub action to generate a release artifact. This is the recommended path for Bazel modules as GitHub automatically generated artifacts may change over time. See https://github.com/bazel-contrib/SIG-rules-authors/issues/11.
    - An example run of the action can be seen here: https://github.com/tweag/rules_nixpkgs/actions/runs/6561135923
    - A draft of the created release can be seen here: https://github.com/tweag/rules_nixpkgs/releases/tag/untagged-30d78a6eeac04526f1da

The next steps are:
- [ ] Review & Merge this PR.
- [x] Delete the draft release.
- [x] Enable the [publish to BCR app](https://github.com/bazel-contrib/publish-to-bcr/tree/main) for this repository.
- [ ] Push a release v0.10.0 tag to the merge commit of this PR.
- [ ] Get rules_nixpkgs_core 0.10.0 merged into the BCR.
- [ ] Add release instructions to rules_nixpkgs. (I'd like to play through the whole process before documenting it)


.github/workflows/release_prep.sh
```diff
@@ -0,0 +1,45 @@
...
+# Set by GH actions, see
+# https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+TAG=${GITHUB_REF_NAME}
+# The prefix is chosen to match what GitHub generates for source archives
+PREFIX="rules_nixpkgs-${TAG:1}"
+ARCHIVE="rules_nixpkgs-${TAG:1}.tar.gz"
+git archive --format=tar.gz --prefix="${PREFIX}/" -o $ARCHIVE ${TAG}
+SHA=$(shasum -a 256 "$ARCHIVE" | awk '{print $1}')
+
+cat << EOF
```
benradf, line 13: If you single quote the heredoc marker:
```suggestion
cat << 'EOF'
```
You can avoid needing to escape the backticks below.

benradf (approved):

.github/workflows/release_prep.sh
aherrmann: There are some variable expansions needed in the heredoc and `<<'EOF'` will keep those verbatim. E.g.
```
$ FOO=foo
$ cat <<EOF
> foo $FOO
> EOF
foo foo
$ cat <<'EOF'
foo $FOO
EOF
foo $FOO
```

aherrmann:
PR to add to BCR https://github.com/bazelbuild/bazel-central-registry/pull/1022

.bcr/core/source.template.json
```diff
@@ -0,0 +1,5 @@
+{
+  "integrity": "",
+  "strip_prefix": "{REPO}-{VERSION}/core",
+  "url": "https://github.com/{OWNER}/{REPO}/releases/download/{TAG}/${REPO}-${VERSION}.tar.gz"
```
prednaz, line 4: why not use `https://github.com/tweag/rules_nixpkgs/archive/refs/tags/v0.10.0.tar.gz` which is available by default?

aherrmann: The short version is that these automatically generated artifacts are not guaranteed to be bit-wise reproducible. But, Bazel uses sha256 to ensure that the downloaded artifact was not tampered with or otherwise damaged. So, any bit-change in the artifact would change the sha256 and cause a failure. So, instead, the recommendation is that Bazel rules produce dedicated release artifacts, which are guaranteed to stay the same.
See also the discussion here: https://github.com/bazel-contrib/SIG-rules-authors/issues/11

prednaz: awesome, thank you!

PR #440, opened 2023-11-07
go: add experiments=nocoverageredesign for go_sdk in 1.20+

lizan:
[rules_go adds experiments=nocoverageredesign](https://github.com/bazelbuild/rules_go/blob/v0.42.0/go/private/sdk.bzl#L486-L491) automatically. rules_nixpkgs doesn't add this, so `bazel coverage` fails with `testing: cannot use -test.coverprofile because test binary was not built with coverage enabled`.


lizan:
@benradf can you take a look on this when you have a moment?

benradf (approved):

PR #450, opened 2023-11-22
nixpkgs_flake_package use `path:` syntax by default

rickvanprim:
Copied from the added documentation:
```md
**IMPORTANT NOTE**: Calling `nix build` copies the entirety of the Nix Flake
into the Nix Store.  When using the `path:` syntax, this means the directory
containing `flake.nix` and any subdirectories.  Without specifying `path:`
Nix may infer that the flake is the Git repository and copy the entire thing.
As a consequence, you may want to isolate your flake from the rest of the
repository to minimize the amount of unnecessary data that gets copied into
the Nix Store whenever the flake is rebuilt.
```

Presently this PR is making `path:` syntax the default behavior as this feels like what people probably intend when invoking Nix via Bazel.  `legacy_path_syntax` is introduced as a flag in case the old behavior is needed.  As far as I'm aware, this should only be breaking if the flake was in a subdirectory of a Git repository and reached upwards (as the whole repository would be copied with the old behavior).  Likewise this doesn't optimize anything if `flake.nix` is located at the root of the repository, but it does open up the potential optimization of isolating the flake from the rest of the repository.

One other consideration is that I gather the `path:` syntax skips/doesn't work with the `evalcache`.  For my impression of how Nix is intended to be used via `rules_nixpkgs`, I don't think this is a big deal (as it's only being built when Bazel detects a change and I'm expecting the flakes to be relatively small using Nix as a package manager not a build system).

The potential cost of copying the whole repository may disappear as part of https://github.com/NixOS/nix/pull/6530.

Please let me know if this change seems reasonable and which direction the default behavior should be.  Secondly, I moved `flake.nix` and `flake.lock` into `//nix` inline with the scoping suggestion I make in the documentation.  I did not move `shell.nix` as I don't believe that has the same issue as flakes and would require touching every invocation of `nix-shell` to specify the path.  Let me know if that's acceptable, if I should revert it, or if you'd also like `shell.nix` moved into `//nix`.

Another option totally different from this would be either a different rule (or just a flag opting in) that copies `flake.nix`, `flake.lock`, and any listed dependencies into an output directory, and then runs `nix build` on that.  This would allow `flake.nix` to still live at the repo root without pulling the entire repo in.


core/nixpkgs.bzl
```diff
@@ -410,26 +410,26 @@ def _nixpkgs_build_and_symlink(repository_ctx, nix_build_cmd, expr_args, build_f
...
+            timeout = timeout,
         )
         output_path = exec_result.stdout.splitlines()[-1]
 
         ssh_path = repository_ctx.which("ssh")
         repository_ctx.report_progress("Creating remote store root")
         exec_result = execute_or_fail(
             repository_ctx,
-            [ssh_path] + [nix_host, "nix-store --add-root ~/rules_nixpkgs_gcroots/{root} -r {path}".format(root = output_path.split('/')[-1], path = output_path) ],
+            [ssh_path] + [nix_host, "nix-store --add-root ~/rules_nixpkgs_gcroots/{root} -r {path}".format(root = output_path.split("/")[-1], path = output_path)],
```
rickvanprim, line 432: Buildifier auto formatting

rickvanprim:
I can't seem to directly add reviewers, so @aherrmann / @layus 🙂

aherrmann (commented):
From my perspective this looks good! Thank you!

@layus WDYT? In particular, is the following a concern?

> One other consideration is that I gather the path: syntax skips/doesn't work with the evalcache. For my impression of how Nix is intended to be used via rules_nixpkgs, I don't think this is a big deal (as it's only being built when Bazel detects a change and I'm expecting the flakes to be relatively small using Nix as a package manager not a build system).

layus:
> @layus WDYT? In particular, is the following a concern?

I think the eval cache is a must. But I also think this should be completely revamped ultimately, that is why I think this can be merged as-is.

Ultimately, I would like to have a flake usable as a nixpkgs repo, and passed to nixpkgs_package. having a nixpkgs_flake_package is odd, but also the best way to move forward for now, so adding this path here makes sense. Way more than having your whole repo copied to the cache, which is unacceptable.

layus (approved):
I am all in for this change right now, in the current state of the repo. See my other comment for the bigger picture and where we should go.

layus:
Or, to be more precise, to have proper eval cache, I think we should declare the flake separately from the nixpkgs_package, like we do for <nixpkgs> inputs. Then using the evaluation cache would come for free.

rickvanprim:
@layus thanks for all the additional context!  Regarding the eval cache thing, I'm just basing that statement on (if I recall correctly) seeing someone saying that `path:` based stuff isn't cached on one of the GitHub Issues while I was trying to figure out relative path flakes.  This surprises me as I'd expect the directory to just get hashed and looked up the same as the Git form, though it can't just pull the hash from Git.

If you can outline how you'd like to see this work, I could try to take a stab at it, if you want.  I feel bad contributing something flawed 😅

layus:
You are not contributing something flawed, hence the fast merge. No, it's rather like we need a bigger picture view, and a plan for future evolutions. Let me draft that somewhere, as discussion sneed to happen with flakes and bzlmod in mind.

joneshf:
> As far as I'm aware, this should only be breaking if the flake was in a subdirectory of a Git repository and reached upwards (as the whole repository would be copied with the old behavior).

There's at least one other scenario where this breaks: if the `flake.nix` is at the root of the repository, and the repository is using the [`core.fsmonitor`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefsmonitor) configuration to enable the [file system monitor](https://git-scm.com/docs/git-fsmonitor--daemon). In this scenario, it causes this bug in nix to manifest: https://github.com/NixOS/nix/issues/11567. For posterity/SEO the error mentions something like:
```
error: file '/path/to/repository/.git/fsmonitor--daemon.ipc' has an unsupported type
```

Thankfully, the `legacy_path_syntax` attribute was added to keep the previous behavior! Folks in this scenario should be able to set it to `True` to not hit that `nix` bug:
```diff
 nixpkgs_flake_package(
     name = "hello",
+    legacy_path_syntax = True,
     nix_flake_file = "//:flake.nix",
     nix_flake_lock_file = "//:flake.lock",
     package = "hello",
 )
```

This is not a big deal, since there's a workaround. Just something to keep in mind in case that attribute is up for removal in the future.

PR #453, opened 2023-11-24
Remove workaround for strict deps checking with nixpkgs JRE

### LINKED ISSUE #299: JavaTestRunner missing dependency warning on MacOS 11
**Describe the bug**
In #273 there are a bunch of changes to the Java toolchain to support using rules_nixpkgs with Bazel 6.
This fixes the main issues, but there is a failure in CI on MacOS while running the Java tests, with an error about needing to explicitly declare a dependency on `@bazel_tools//tools/jdk:TestRunner`, with error message recommending running `buildozer 'add deps @bazel_tools//tools/jdk:TestRunner' //tests:java-test`. Explicitly adding the dependency does not solve the problem and results in the same issue.
The temporary workaround for now is to setting the flag `--experimental_strict_java_deps=warn`, but we would like to remove this flag and have the Java tests work.

**To Reproduce**
In `flake.nix`, use nixpkgs version 22.11, and switch `bazel_5` to `bazel_6` in the build inputs.
In `.bazelrc`, remove the line `build --experimental_strict_java_deps=warn`.
Run the Java toolchain tests on MacOS 11:
```console
cd toolchains/java
bazel test '//...'
```

**Expected behavior**
We should be able to remove the flag `--experimental_strict_java_deps=warn`, run the Java toolchain tests, and not have any failures

**Environment**
 - OS name + version: MacOS 11 (in CI)
 - Version of the code: f3d68f3ea2e1075ccddb4aa1013a477b9508a5d6

**Additional context**
My guess is that this is related to [how Bazel 6 is packaged ](https://github.com/NixOS/nixpkgs/tree/master/pkgs/development/tools/build-managers/bazel/bazel_6) in nixpkgs, where there are some particular patches to the the Java toolchain.

avdv:
The JRE we get from nixpkgs is now recent enough and strict deps checking works properly.

Fixes #299

### DIFF
--- a/.bazelrc.java
+++ b/.bazelrc.java
@@ -9,10 +9,4 @@ build --java_language_version=11
 build --tool_java_runtime_version=nixpkgs_java_11
 build --tool_java_language_version=11
 
-# The following lines provide a workaround for a test failure in on the Java
-# toolchain on MacOS 11, documented here: https://github.com/tweag/rules_nixpkgs/issues/299.
-# They should be removed as soon as that issue has been resolved.
-build --enable_platform_specific_config
-build:macos --experimental_strict_java_deps=warn
-
 # vim: ft=conf


aherrmann (approved):
:tada:

PR #465, opened 2023-12-11
`nixpkgs_flake_package()` supplies `--experimental-features 'nix-command flakes'` in case someone's local `nix.conf` hasn't enabled them.

rickvanprim:
@layus does this seem reasonable?  I'm not aware of any downsides for unconditionally passing these flags.

### DIFF
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -790,7 +790,7 @@ def _nixpkgs_flake_package_impl(repository_ctx):
         extra_msg = "See: https://nixos.org/nix/",
     )
 
-    _nixpkgs_build_and_symlink(repository_ctx, [nix_path, "build"], expr_args, build_file_content)
+    _nixpkgs_build_and_symlink(repository_ctx, [nix_path, "--extra-experimental-features", "nix-command flakes", "build"], expr_args, build_file_content)
 
 _nixpkgs_flake_package = repository_rule(
     implementation = _nixpkgs_flake_package_impl,


core/nixpkgs.bzl
```diff
@@ -790,7 +790,7 @@ def _nixpkgs_flake_package_impl(repository_ctx):
         extra_msg = "See: https://nixos.org/nix/",
     )
 
-    _nixpkgs_build_and_symlink(repository_ctx, [nix_path, "build"], expr_args, build_file_content)
+    _nixpkgs_build_and_symlink(repository_ctx, [nix_path, "--experimental-features", "nix-command flakes", "build"], expr_args, build_file_content)
```
layus, line 793:
```suggestion
    _nixpkgs_build_and_symlink(repository_ctx, [nix_path, "--extra-experimental-features", "nix-command flakes", "build"], expr_args, build_file_content)
```

layus (changes requested):
I think you need --extra-experimental-features to avoid wiping the features already present in the user config. Otherwise, yes, this is a _very_ good idea :100:

core/nixpkgs.bzl
rickvanprim: Good idea!  Done 🙂

rickvanprim:
One more thought I have while I'm in here, `is_supported_platform()` is currently:
```bazel
def is_supported_platform(repository_ctx):
    return repository_ctx.which("nix-build") != None
```
Should this check to checking `nix` at some point?  And/or should `nixkpkgs_flake_package()` check `nix` and the other rules continue checking `nix-build`?

aherrmann:
The existence of `nix-build` is just used as a proxy for whether Nix is installed. With modern versions of Nix it could equivalently check for `nix`, I think. But, I don't think we need to conflate these in this PR.

benradf (approved):

layus (approved):

PR #475, opened 2024-01-15
Add SystemConfiguration to the CC wrapper

bradzacher:
Noticed that some rust builds were failing with linker errors after we added the `reqwest` crate. Turns out this was missing from the wrapper - patching this in locally fixed it.


benradf (approved):
Thanks for submitting this fix.

PR #492, opened 2024-03-29
add `compatibility_level` to `core/MODULE.bazel` before publish to BCR

prednaz:
[`add_module.py`](https://github.com/bazelbuild/bazel-central-registry/tree/main/tools#add_modulepy) shouts the following at me.

```
BcrValidationResult.FAILED: Checked in MODULE.bazel file doesn't match the one in the extracted and patched sources.
Please fix the MODULE.bazel file or you can add the following patch to rules_nixpkgs_core@0.11.0:
    --- MODULE.bazel
    +++ MODULE.bazel
    @@ -1,15 +1,7 @@
     module(
         name = "rules_nixpkgs_core",
         version = "0.11.0",
    +    compatibility_level = 1,
     )
```

now i am worried that `compatibility_level` is required in `MODULE.bazel` now. does anybody know more?

### DIFF
--- a/core/MODULE.bazel
+++ b/core/MODULE.bazel
@@ -1,6 +1,7 @@
 module(
     name = "rules_nixpkgs_core",
     version = "0.11.0",
+    compatibility_level = 1,
 )
 
 bazel_dep(name = "platforms", version = "0.0.4")


cgrindel:
The `add_module.py` is wrong. It states [that the compatibility_level default is 1](https://github.com/bazelbuild/bazel-central-registry/blob/e586cc239bab96ada2d7bbf7058cb80b3e5a3da0/tools/add_module.py#L82-L83). It is not. [It is 0](https://bazel.build/rules/lib/globals/module#module). We should not set this until we absolutely must.

cgrindel:
I suggest that we close this PR.

PR #497, opened 2024-04-08
make `rules_nixpkgs_java` depend on `remote_java_tools` properly

prednaz:
this fixes the issue that users of `rules_nixpkgs_java` need to depend on a dependency of `rules_nixpkgs_java`, that is on `remote_java_tools`. examples are [`8e44cce`](https://github.com/tweag/rules_nixpkgs/pull/497/commits/8e44ccea8442bfb3b3a3386494a0cf90f088c295).


prednaz:
@layus pointed out that this fix has also long been available from upstream, that is from [`rules_java`](https://github.com/bazelbuild/rules_java/blob/master/toolchains/default_java_toolchain.bzl). is there more? can we deduplicate something here?

benradf (approved):

PR #502, opened 2024-04-11
update `nixpkgs` in `examples/toolchains`

prednaz:


.github/workflows/workflow.yaml
```diff
@@ -81,7 +81,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: cachix/install-nix-action@v24
         with:
-          nix_path: nixpkgs=./nixpkgs.nix
+          nix_path: nixpkgs=../nixpkgs.nix
```
prednaz, line 84: why is this necessary? i was getting [an error message i do not understand](https://github.com/tweag/rules_nixpkgs/actions/runs/8647896527/job/23710437150) before i adjusted this. the reported line `(import <nixpkgs> {}).bashInteractive` does not occur anywhere in the entire repository.

prednaz: this line is problematic for the rust example because it needs its own `nixpkgs.nix`.

PR #526, opened 2024-05-09
Catch case in which find_children is called on an empty dir

numine777:
In our uses, we have found that nix store package corruption can cause a nasty error if `find_children` is called on an empty directory. Specifically, the output of a subprocess call using the find command for `find_children` actually returns `['']` if called on an empty directory. This causes a nasty bug in the for loop designed to "build a forest of symlinks to the nix store" in which symlink gets called incorrectly and causes a tricky bazel error to debug. This change should allow users to get a more helpful error message for debugging purposes.

### DIFF
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -469,7 +469,7 @@ def _nixpkgs_build_and_symlink(repository_ctx, nix_build_cmd, expr_args, build_f
     # Nix store.
     for target in find_children(repository_ctx, output_path):
         if target == '':
-            fail("It appears that you nix store may have a corrupt entry. Found empty directory: {}".format(output_path))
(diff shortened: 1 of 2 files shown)


benradf:
Nice find, that's quite a subtle bug.

I think we should fix `find_children` so that it returns `[]` for an empty directory. The question then is how would a corrupted store / empty directory manifest in your use case? I guess you'd get a `nixpkgs_package` repository with no symlinks in it, which would cause errors further down the line.

So having the check you added is worthwhile. I've pushed a commit with my suggestions — let me know what you think.

benradf (approved):
@numine777 Feel free to merge this if you're happy with my changes.

numine777:
> @numine777 Feel free to merge this if you're happy with my changes.

I'm not sure that I have the authority, but happy to if I'm granted permissions

Issue #574, opened 2024-08-14
Unfilled template value in Bazel v7.3.0 and v8 preview releases


Silic0nS0ldier:
**Describe the bug**

As of https://github.com/bazelbuild/bazel/commit/9e913fbfa2e3930742fc30dfb60ac5e2694c70cf (was backported to 7.3.0 in https://github.com/bazelbuild/bazel/commit/6dda687f24a1e6b3000ab3ae01742d2b6829cb18) a replacement value for `%{extra_flags_per_feature}` must be specified (done in https://github.com/tweag/rules_nixpkgs/blob/343aaa474deb8b29ac717550be322b13a4d50077/toolchains/cc/cc.bzl#L187) in order for a valid `BUILD` file to be produced.

**To Reproduce**

Use a Nix CC toolchain in Bazel 7.3.0 or `8.0.0-pre.20240807.1`.

**Expected behavior**

Build passes.

**Environment**
 - OS name + version: macOS
 - Version of the code: v0.11.1 (with patches)

**Additional context**
Nope, this is just a breakage caused by a new value being required in Bazel.

PR #575, opened 2024-08-23
Add extra_flags_per_feature to cc toolchain

hofbi:
Closes https://github.com/tweag/rules_nixpkgs/issues/574

### DIFF
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -298,7 +298,7 @@ pkgs.runCommand "bazel-${cc.orignalName or cc.name}-toolchain"
         if [[ -x ${cc}/bin/ld.gold ]]; then echo True; else echo False; fi
       )
     )
-    COVERAGE_LINK_FLAGS=()
+    EXTRA_FLAGS_PER_FEATURE=()
     IS_CLANG=(
       ${if cc.isClang then "True" else "False"}
     )
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -298,6 +298,7 @@ pkgs.runCommand "bazel-${cc.orignalName or cc.name}-toolchain"
         if [[ -x ${cc}/bin/ld.gold ]]; then echo True; else echo False; fi
       )
     )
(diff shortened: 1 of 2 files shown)


toolchains/cc/cc.nix
```diff
@@ -298,6 +298,7 @@ pkgs.runCommand "bazel-${cc.orignalName or cc.name}-toolchain"
         if [[ -x ${cc}/bin/ld.gold ]]; then echo True; else echo False; fi
       )
     )
+    COVERAGE_LINK_FLAGS=()
```
aherrmann, line 301: why is this needed? It overrides the [set of flags defined above](https://github.com/tweag/rules_nixpkgs/pull/575/files#diff-ebda446a0ba42acd64ea7fe331d17801b2c3f88c96e3fe1d785f95345404523dR288).

hofbi:
```suggestion
    EXTRA_FLAGS_PER_FEATURE=()
```

aherrmann (approved):
Thank you!

toolchains/cc/cc.nix
```diff
@@ -298,6 +298,7 @@ pkgs.runCommand "bazel-${cc.orignalName or cc.name}-toolchain"
         if [[ -x ${cc}/bin/ld.gold ]]; then echo True; else echo False; fi
       )
     )
+    EXTRA_FLAGS_PER_FEATURE=()
```
Silic0nS0ldier, line 301: Discovered while testing that Bazel expects a `dict` here. Noted at https://github.com/tweag/rules_nixpkgs/issues/574#issuecomment-2314438748.

`tool_paths` is an example of a `dict` whose value is set here, the problem is that is a bash associative array (it is represented as a dict as that is the only way it can be).

It may be necessary to handle the case where `repr` returns `[]` in `toolchains/cc/cc.bzl` to fix this.

aherrmann: See also https://github.com/tweag/rules_nixpkgs/pull/576

PR #576, opened 2024-08-26
Support empty extra_flags_per_feature

hofbi:
Used the wrong serialization in https://github.com/tweag/rules_nixpkgs/pull/575. Since this is not really used yet, we set always set an empty dict for now.

Closes https://github.com/tweag/rules_nixpkgs/issues/578

### DIFF
--- a/toolchains/cc/cc.bzl
+++ b/toolchains/cc/cc.bzl
@@ -221,7 +221,7 @@ def _nixpkgs_cc_toolchain_config_impl(repository_ctx):
             "%{coverage_compile_flags}": get_starlark_list(info.coverage_compile_flags),
             "%{coverage_link_flags}": get_starlark_list(info.coverage_link_flags),
             "%{supports_start_end_lib}": repr(info.supports_start_end_lib),
-            "%{extra_flags_per_feature}": repr(info.extra_flags_per_feature),
+            "%{extra_flags_per_feature}": repr(dict(info.extra_flags_per_feature)),
             "%{conly_flags}": get_starlark_list(info.conly_flags),
         },
     )
--- a/toolchains/cc/cc.bzl
+++ b/toolchains/cc/cc.bzl
@@ -78,7 +78,6 @@ def _parse_cc_toolchain_info(content, filename):
         "COVERAGE_COMPILE_FLAGS",
         "COVERAGE_LINK_FLAGS",
         "SUPPORTS_START_END_LIB",
-        "EXTRA_FLAGS_PER_FEATURE",
         "IS_CLANG",
         "CONLY_FLAGS",
     ])
@@ -119,7 +118,6 @@ def _parse_cc_toolchain_info(content, filename):
         coverage_compile_flags = info["COVERAGE_COMPILE_FLAGS"],
         coverage_link_flags = info["COVERAGE_LINK_FLAGS"],
         supports_start_end_lib = info["SUPPORTS_START_END_LIB"] == ["True"],
-        extra_flags_per_feature = info["EXTRA_FLAGS_PER_FEATURE"],
         is_clang = info["IS_CLANG"] == ["True"],
         conly_flags = info["CONLY_FLAGS"],
     )
@@ -221,7 +219,7 @@ def _nixpkgs_cc_toolchain_config_impl(repository_ctx):
             "%{coverage_compile_flags}": get_starlark_list(info.coverage_compile_flags),
             "%{coverage_link_flags}": get_starlark_list(info.coverage_link_flags),
             "%{supports_start_end_lib}": repr(info.supports_start_end_lib),
-            "%{extra_flags_per_feature}": repr(dict(info.extra_flags_per_feature)),
+            "%{extra_flags_per_feature}": "{}",
             "%{conly_flags}": get_starlark_list(info.conly_flags),
         },
     )
--- a/toolchains/cc/cc.nix
+++ b/toolchains/cc/cc.nix
@@ -298,7 +298,6 @@ pkgs.runCommand "bazel-${cc.orignalName or cc.name}-toolchain"
         if [[ -x ${cc}/bin/ld.gold ]]; then echo True; else echo False; fi
       )
     )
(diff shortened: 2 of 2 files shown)


jtszalay:
I tried to use this (as I was still getting an error after PR #575:

```
ERROR: ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel:154:31: syntax error at ',': expected expression
ERROR: ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel: no such target '@@nixpkgs_config_cc//:cc-compiler-k8': target 'cc-compiler-k8' not declared in package '' defined by ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel
ERROR: ../bazel-user-root/output/external/rules_go~/BUILD.bazel:86:17: Target '@@rules_go~//:cgo_context_data' depends on toolchain '@@nixpkgs_config_cc//:cc-compiler-k8', which cannot be found: no such target '@@nixpkgs_config_cc//:cc-compiler-k8': target 'cc-compiler-k8' not declared in package '' defined by ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel'
```

aherrmann:
Interesting, the [template built into Bazel](https://github.com/bazelbuild/bazel/blob/41da8a42510f6d9b6ea1937077997e8019d75eb9/tools/cpp/BUILD.tpl#L121C28-L121C55) includes this
```
    extra_flags_per_feature = %{extra_flags_per_feature},
```
other fields that should be set via `get_starlark_list` look like this instead
```
    coverage_link_flags = [%{coverage_link_flags}],
```

FWIW, [Bazel itself also uses `repr`](https://github.com/bazelbuild/bazel/blob/41da8a42510f6d9b6ea1937077997e8019d75eb9/tools/cpp/unix_cc_configure.bzl#L732). @hofbi In which way does it fail without this PR?

hofbi:
Without this PR, I get the same error as reported by @jtszalay 

```
ERROR: ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel:154:31: syntax error at ',': expected expression
ERROR: ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel: no such target '@@nixpkgs_config_cc//:cc-compiler-k8': target 'cc-compiler-k8' not declared in package '' defined by ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel
ERROR: ../bazel-user-root/output/external/rules_go~/BUILD.bazel:86:17: Target '@@rules_go~//:cgo_context_data' depends on toolchain '@@nixpkgs_config_cc//:cc-compiler-k8', which cannot be found: no such target '@@nixpkgs_config_cc//:cc-compiler-k8': target 'cc-compiler-k8' not declared in package '' defined by ../bazel-user-root/output/external/nixpkgs_config_cc/BUILD.bazel'
```

I will try to fully reproduce and test with a local_override but might only get to to next week.

avdv:
> Interesting, the [template built into Bazel](https://github.com/bazelbuild/bazel/blob/41da8a42510f6d9b6ea1937077997e8019d75eb9/tools/cpp/BUILD.tpl#L121C28-L121C55) includes this
> 
> ```
>     extra_flags_per_feature = %{extra_flags_per_feature},
> ```

Yes, but In Bazel, the `extra_flags_per_feature` variable is a dict (see https://github.com/bazelbuild/bazel/blob/e88a7d982411e3d6bbfb658b88e76944ea5ed720/tools/cpp/unix_cc_configure.bzl#L559C5-L559C33) and using `repr` is just writing that dict to the BUILD file.

So, the `EXTRA_FLAGS_PER_FEATURE` variable in cc.nix should be an associative array and be written to the CC_TOOLCHAIN_INFO file in a form that makes it possible to decode it as a dict again. Arguably, since the variable never has a meaningful value currently, always writing `{}` would be the easiest thing to fix this.

aherrmann:
Thank you @avdv, good catch!

> Arguably, since the variable never has a meaningful value currently, always writing `{}` would be the easiest thing to fix this.

Yes, that's an easy quick fix.

> So, the `EXTRA_FLAGS_PER_FEATURE` variable in cc.nix should be an associative array and be written to the CC_TOOLCHAIN_INFO file in a form that makes it possible to decode it as a dict again.

This may be a good motivation to finally replace the `CC_TOOLCHAIN_INFO` format by json. This format was introduced before [Bazel gained `json.decode`](https://bazel.build/rules/lib/core/json#decode).

toolchains/cc/cc.bzl
```diff
@@ -221,7 +221,7 @@ def _nixpkgs_cc_toolchain_config_impl(repository_ctx):
             "%{coverage_compile_flags}": get_starlark_list(info.coverage_compile_flags),
             "%{coverage_link_flags}": get_starlark_list(info.coverage_link_flags),
             "%{supports_start_end_lib}": repr(info.supports_start_end_lib),
-            "%{extra_flags_per_feature}": repr(info.extra_flags_per_feature),
+            "%{extra_flags_per_feature}": repr(dict(info.extra_flags_per_feature)),
```
avdv, line 224: `dict()` expects an iterable of pairs as its first argument. `info.extra_flags_per_feature` is a list of strings. This will only work out if this list is empty. So, this is a bit complicated way of always writing out `{}`.

Is there a way forward for the `EXTRA_FLAGS_PER_FEATURE` in `cc.nix` to be useful? Otherwise I'd rather remove it until there is a use case for it.

hofbi: Would you suggest we simply do

```suggestion
            "%{extra_flags_per_feature}": repr({}),
```

and remove the rest until there is a concrete usecase?

avdv: Yes, even just using `"{}"` would do.

hofbi: Done

avdv (approved):
Thank you @hofbi , LGTM!

avdv (approved):

hofbi:
@mergifyio update

PR #577, opened 2024-08-27
nixpkgs_cc_toolchain_config: patch dynamic linker feature on macOS

malt3:
This is a workaround until
https://github.com/bazelbuild/bazel/pull/23438 is merged and available.


malt3:
Not a correct fix.

PR #581, opened 2024-09-03
Re-enable remote nix tests

### LINKED ISSUE #564: nix remote jobs are not working correctly on CI
**Describe the bug**

I noticed that the nix remote CI jobs (`withNixRemote = true`) actually don't run remotely since the `NIX_REMOTE_ENABLED` variable which is checked here:
https://github.com/tweag/rules_nixpkgs/blob/c1afab1763a238bad08a76bb91a12eb9ac2acecc/.github/workflows/workflow.yaml#L60-L63
is actually set to `matrix.withNixRemote && matrix.os == 'ubuntu-latest'` verbatim, so the condition is never true.

After fixing that problem (see https://github.com/tweag/rules_nixpkgs/commit/5046b7dca09b73824a7bd1fb4f850c1b6f11e9e9), the jobs fail with:
```
       Return code: 1
      Error output: 
        > warning: unknown setting 'always-allow-substitutes'
        > ssh: connect to host localhost port 2222: Connection refused
        > error: failed to start SSH connection to 'nix-server'
```
See https://github.com/tweag/rules_nixpkgs/actions/runs/9875397903/job/27271997561

**To Reproduce**

**Expected behavior**
CI jobs should be able to use the nix server that is started inside a Docker container.

**Environment**
 - OS name + version: ubuntu-latest
 - Version of the code: 

**Additional context**
I am going to disable the jobs for now.  @z8v could you have a look to see what's going on?

avdv:
While debugging some other Docker image I noticed that the container did not really start and searching for that error it turned there would be a really simple fix.

The sshd process did not start:
```
Missing privilege separation directory: /run/sshd
```

Ensure to create this directory and make sshd log messages to stderr.

Fixes #564

### DIFF
--- a/.github/nix-server/Dockerfile
+++ b/.github/nix-server/Dockerfile
@@ -17,8 +17,11 @@ COPY .github/nix-server/keys .
 RUN cat ci.pub > $HOME/.ssh/authorized_keys
 
 RUN echo 'PermitRootLogin yes' >> /etc/ssh/sshd_config && \
+    mkdir -p /run/sshd && \
     echo ". /root/.nix-profile/etc/profile.d/nix.sh" >> $HOME/.bashrc && \
     ln -sf /root/.nix-profile/bin/nix-store /usr/bin/ && \
     ln -sf /root/.nix-profile/bin/nix-daemon /usr/bin/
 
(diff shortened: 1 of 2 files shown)


benradf (approved):
Looks good to me :+1:

PR #601, opened 2024-10-08
Add a way to access the output_path in generated BUILD files

layus:
This is really useful to avoid Bazel seeing the files themselves considering they are already in the nix store. No point in copying them in the bazel cache

### DIFF
--- a/core/nixpkgs.bzl
+++ b/core/nixpkgs.bzl
@@ -478,7 +478,7 @@ def _nixpkgs_build_and_symlink(repository_ctx, nix_build_cmd, expr_args, build_f
     # provided by `build_file` or `build_file_content`.
     if not repository_ctx.path("BUILD").exists and not repository_ctx.path("BUILD.bazel").exists:
         if build_file_content:
-            repository_ctx.file("BUILD", content = build_file_content)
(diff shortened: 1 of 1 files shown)


core/nixpkgs.bzl
```diff
@@ -478,7 +478,7 @@ def _nixpkgs_build_and_symlink(repository_ctx, nix_build_cmd, expr_args, build_f
     # provided by `build_file` or `build_file_content`.
     if not repository_ctx.path("BUILD").exists and not repository_ctx.path("BUILD.bazel").exists:
         if build_file_content:
-            repository_ctx.file("BUILD", content = build_file_content)
+            repository_ctx.file("BUILD", content = build_file_content.replace("{{NIX_STORE_PATH}}", output_path))
```
aherrmann, line 481: Nice, that could potentially have simplified things like the [Python toolchain](https://github.com/tweag/rules_nixpkgs/blob/567b8713dd10e467765624683aa41a44b7110a4e/toolchains/python/python.bzl#L92).

I guess the same could be done with the `build_file` attribute using `repository_ctx.template` for consistency.

This behavior should also be documented in the relevant repo macros / module extension tags.

PR #621, opened 2024-11-29
remove nixpkgs_cc_configure_deprecated

malt3:
`nixpkgs_cc_configure_deprecated` depended on Bazel's cc_autoconf_impl, which was removed from @bazel_tools and moved into a private package in rules_cc in Bazel 8.

See also: bazelbuild/bazel#23809


avdv (approved):
:+1:

PR #639, opened 2025-01-28
refactor darwin CC toolchain to work with nixpkgs 24.11

### LINKED ISSUE #623: can't use `nixpkgs_cc_configure` on darwin with nixpkgs 24.11
**Describe the bug**
It looks like the [changes](https://discourse.nixos.org/t/the-darwin-sdks-have-been-updated/55295/27) made to improve the Darwin SDK situation have inadvertently broken using `nixpkgs_cc_configure` on darwin systems.

```
  Cannot build Nix derivation for package '@nixpkgs_config_cc_info'.
    Command: "/run/current-system/sw/bin/nix-build" "-I" "nixpkgs=/private/var/tmp/_bazel_flurie/329cba5e783b87d02ca74eb12aa46836/external/nixpkgs/nixpkgs" "/private/var/tmp/_bazel_flurie/329cba5e783b87d02ca74eb12aa46836/external/nixpkgs_config_cc_info/external/rules_nixpkgs_cc/cc.nix" "-A" "" "--out-link" "bazel-support/nix-out-link" "--argstr" "ccType" "ccTypeDefault" "--argstr" "ccLang" "c++" "--argstr" "ccStd" "c++2b"
    Return code: 1
    Error output: 
      > error:
      >        … while evaluating the 'name' attribute of a derivation
      > 
      >        … while evaluating a branch condition
      >          at /private/var/tmp/_bazel_flurie/329cba5e783b87d02ca74eb12aa46836/external/nixpkgs/lib/strings.nix:2627:3:
      >          2626|   # First detect the common case of already valid strings, to speed those up
      >          2627|   if stringLength string <= 207 && okRegex string != null
      >              |   ^
      >          2628|   then unsafeDiscardStringContext string
      > 
      >        … in the left operand of the AND (&&) operator
      >          at /private/var/tmp/_bazel_flurie/329cba5e783b87d02ca74eb12aa46836/external/nixpkgs/lib/strings.nix:2627:33:
      >          2626|   # First detect the common case of already valid strings, to speed those up
      >          2627|   if stringLength string <= 207 && okRegex string != null
      >              |                                 ^
      >          2628|   then unsafeDiscardStringContext string
      > 
      >        (stack trace truncated; use '--show-trace' to show the full, detailed trace)
      > 
      >        error: function 'anonymous lambda' called with unexpected argument 'postLinkSignHook'
      >        at /private/var/tmp/_bazel_flurie/329cba5e783b87d02ca74eb12aa46836/external/nixpkgs/pkgs/build-support/bintools-wrapper/default.nix:8:1:
      >             7|
      >             8| { name ? ""
      >              | ^
      >             9| , lib
```

**To Reproduce**
For now you can pull https://github.com/flurie/aoc-2024 and [repoint](https://github.com/flurie/aoc-2024/blob/main/WORKSPACE.bazel#L28) the cc to 24.11. I can produce a min repro if I have time and if it's needed.

**Expected behavior**
The rework looks pretty extensive, so I don't know that I expect it to work out of the box, but it does look like `postLinkSignHook` is no longer an overrideable scope, and it also looks like the patch has been incorporated upstream, so there's a chance it will work after removing that override.

**Environment**
 - OS name + version: maOS 14.5
 - Version of the code: master

**Additional context**

malt3:
Fixes #623

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ nixpkgs_package(
 <pre>
 nixpkgs_cc_configure(<a href="#nixpkgs_cc_configure-name">name</a>, <a href="#nixpkgs_cc_configure-attribute_path">attribute_path</a>, <a href="#nixpkgs_cc_configure-nix_file">nix_file</a>, <a href="#nixpkgs_cc_configure-nix_file_content">nix_file_content</a>, <a href="#nixpkgs_cc_configure-nix_file_deps">nix_file_deps</a>, <a href="#nixpkgs_cc_configure-repositories">repositories</a>,
                      <a href="#nixpkgs_cc_configure-repository">repository</a>, <a href="#nixpkgs_cc_configure-nixopts">nixopts</a>, <a href="#nixpkgs_cc_configure-quiet">quiet</a>, <a href="#nixpkgs_cc_configure-fail_not_supported">fail_not_supported</a>, <a href="#nixpkgs_cc_configure-exec_constraints">exec_constraints</a>,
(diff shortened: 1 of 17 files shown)


tshaynik (approved):
Looks great!

PR #644, opened 2025-03-27
fix remote argument to nixpkgs_git_repository

NilsIrl:
remote argument was previously unused


malt3 (approved):
Thank you for this PR! It looks good to me.

PR #664, opened 2025-08-25
Add run_nix_shell

### LINKED ISSUE #615: Add an easy way to use packages inside run_shell
As a developer I want to invoke external buildsystems/tools that may do execs themselves. For this I would need a proper PATH with all nixpkgs I want configured. 

Something like run_shell but with another argument where either one label to a nix_shell_env (bad name i know) which contains all packages, or the list of packages are supplied and fitted into the run_shell environment. Bonus points if include paths etc are correct set

Hacking into nix-shell implementation to get a shell file I can use for entering. Doing it manually by including all packages manually, try rules_sh (sadly won't work in my case because of bzlmod dependency trees).

**Additional context**
Basically I want to use rules_nixpkgs as bazel sandbox provider

jlesquembre:
Fixes #615


core/nixpkgs.bzl
```diff
@@ -884,3 +884,92 @@ def nixpkgs_flake_package(
...
+
+    ctx.actions.run_shell(
+      inputs = [nix_env],
+      outputs = [out_file],
+      progress_message = "Running nix shell...",
+      arguments = [out_file.path],
+      command = """
+        source {}
+        {}
+        """.format(nix_env.path, ctx.attr.cmd, out_file.path)
```
layus: This is strange, you are passing three arguments to format, while there are only two placeholders.

core/nixpkgs.bzl
```diff
@@ -903,3 +884,92 @@ def nixpkgs_flake_package(
...
+    )
+    return [DefaultInfo(files = depset([out_file]))]
+
+run_nix_shell = rule(
+    implementation = _run_nix_shell_impl,
+    attrs = {
+        "output": attr.string(
```
malt3:
```suggestion
        "outputs": attr.output_list(
```

I think an [output list](https://bazel.build/rules/lib/toplevel/attr#output_list) would give us a bit more flexibility. This way, users can define multiple outputs for a single action. This also avoids the need to `ctx.actions.declare_file`. Instead, the file is available through `ctx.outputs.outputs`.

malt3 (commented):
Looks pretty good to me. I think this would be a little nicer using an output_list.

jlesquembre:
@tweag/rules_nixpkgs_reviewers Rebased onto the latest master branch, which resolved the CI issues. I believe that was the only thing blocking this PR.

avdv (approved):
LGTM, thank you!

(I have just wired up the missing documenation and regenerated the README)

PR #711, opened 2026-04-09
Add macos-15 runner

tweag-ev-ak:
Macos-26 is the last one that will support Intel so we should start using ARM builders.


avdv (approved):
Looks good, thank you!

PR #742, opened 2026-08-04
Update to Bazel 7

vreuter:


core/private/cc_toolchain/linux_cc_wrapper.sh.tpl
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
avdv: Usually this is what we try to avoid: hard coding the bash interpreter; which means this does not work on NixOS. (there is no /bin/bash)

Optimally we should find a different solution. Maybe we can insert the bash path in the shebang when expanding this template instead?

.github/actions/install-bash-on-macOS/action.yaml
```diff
@@ -0,0 +1,13 @@
...
+  using: composite
+  steps:
+    - shell: bash
+      if: runner.os == 'macOS'
+      run: |
+        # macOS ships with an older bash, but nixpkgs 25.05 requires
+        # Bash 5+. Install new bash via Homebrew and set NIX_BUILD_SHELL so
+        # nix-shell uses it for environment setup.
+        brew install bash
+        echo "NIX_BUILD_SHELL=$(brew --prefix)/bin/bash" >> "$GITHUB_ENV"
```
avdv: Can we use nix instead of homebrew? 

And is the comment about nixpkgs 25.05 actually right? Maybe this should read 26.05?

jlesquembre: Am I correct in thinking the only reason to install Bash is to fix the CI error on `darwin-intel` (`x86_64-darwin` + nixpkgs 11.26)? Since nixpkgs has dropped support for `x86_64-darwin` , we should consider dropping support for it as well. Doing so would make installing Bash on macOS (and this action) unnecessary

vreuter: > And is the comment about nixpkgs 25.05 actually right? Maybe this should read 26.05?

@avdv Yes, for these changes I'm using 25.05, so I had the comment to match that. IDK, should we bump already to 26.05?

core/private/cc_toolchain/linux_cc_wrapper.sh.tpl
vreuter: :ok_hand:

.github/actions/install-bash-on-macOS/action.yaml
vreuter: > Am I correct in thinking the only reason to install Bash is to fix the CI error on `darwin-intel` (`x86_64-darwin` + nixpkgs 11.26)? Since nixpkgs has dropped support for `x86_64-darwin` , we should consider dropping support for it as well. Doing so would make installing Bash on macOS (and this action) unnecessary

@jlesquembre yes, this addition is just to fix the CI, but I will need to check if it was also on non-Intel macOS, or just Intel-based macOS, where this was necessary.

vreuter: @avdv WDYT about dropping macOS Intel support as suggested by @jlesquembre ?

avdv: If it is not too much work, I think we should keep macos-intel for now.  That system is going to be unsupported in nixos 26.11, but that's still a few months away (until we decide to upgrade).

vreuter: The use sites for the action have been removed, so the action itself has been removed now, too.

.github/actions/remove-homebrew-bazel/action.yaml
```diff
@@ -0,0 +1,17 @@
...
+      run: |
+        # GitHub's macOS runner images ship Bazel pre-installed via Homebrew,
+        # and Homebrew's `bazel` is actually bazelisk. `/opt/homebrew/bin` is
+        # not stripped by `nix-shell --pure` on macOS and ends up ahead of the
+        # Nix store on PATH, so `bazel` resolves to the Homebrew bazelisk
+        # instead of the Nix-provided bazel_7. Bazelisk then downloads whatever
+        # version it resolves (e.g. the latest stable) rather than using the
+        # Bazel 7 we pin via Nix. Remove it so only the Nix bazel remains.
+        brew uninstall --ignore-dependencies bazelisk bazel 2>/dev/null || true
+        rm -f "$(brew --prefix)/bin/bazel" "$(brew --prefix)/bin/bazelisk"
```
avdv: This action should also no longer be necessary, it is fixing a problem which was introduced fixing the bash problem with `NIX_BUILD_SHELL=$(brew --prefix)/bin/bash"`:

```console
$ nix-shell --pure --run 'type -a fzf'
fzf is /nix/store/cabxqwq1964bglm63scbjpk64v2fgh6k-fzf-0.55.0/bin/fzf

$ NIX_BUILD_SHELL=$(brew --prefix)/bin/bash nix-shell --pure --run 'type -a fzf'
fzf is /opt/homebrew/bin/fzf
fzf is /nix/store/cabxqwq1964bglm63scbjpk64v2fgh6k-fzf-0.55.0/bin/fzf
```
So using the homebrew bash has the ill effect of adding homebrew's bin directory to the PATH.

core/private/cc_toolchain/linux_cc_wrapper.sh.tpl
vreuter: Thanks @avdv , this is addressed here: da607a2f662958e5ced5571865499fa78bfa280e

.github/actions/remove-homebrew-bazel/action.yaml
avdv: I reverted the relevant commits, let's see if things still work...

vreuter: I was trying to find where this got added; do you have the commit?

vreuter: ah nevermind, got it

avdv (approved):
@vreuter Thank you for pushing this through!

PR #745, opened 2026-08-21
CI: fix nix SSL error

jlesquembre:
Also switch from `DeterminateSystems/nix-installer-action` to `cachix/install-nix-action`,  wince we want to use upstream Nix anyway


avdv (approved):
LGTM, thank you!

PR #750, opened 2026-08-27
Add .direnv/ to gitignore

tshaynik:
We are already using direnv, so we should ignore the generated .direnv directory which are not necessary to have in version control.


avdv (approved):
Thanks!

PR #754, opened 2026-08-28
Update nixpkgs and related rules

jlesquembre:
Updating nixpkgs requires updating rules_go and rules_rust, as well as migrating test dependencies from poetry2nix to uv2nix.


toolchains/go/MODULE.bazel
```diff
@@ -8,6 +8,6 @@ local_path_override(
    module_name = "rules_nixpkgs_core",
    path = "../../core",
 )
-bazel_dep(name = "rules_go", repo_name = "io_bazel_rules_go", version = "0.39.1")
+bazel_dep(name = "rules_go", repo_name = "io_bazel_rules_go", version = "0.63.0")
```
avdv, line 11: Is it really required to bump rules_go to the latest version? 

Note, it is OK for the test module, but for the published modules we want to be much more conservative. Meaning we should choose the minimal version required to make it work. Downstream users are always able to use a newer version, but it's not easily possible to downgrade a dependency.

Issue #749, opened 2026-08-27
Prepare release 0.14.0


avdv:
- [x] Read through this process in its entirety so you understand it.
- [x] Create and checkout a new release preparation branch, named
      `release-<major>.<minor>.<patch>`.
- [x] If the minimal Bazel version has changed:
  - [x] update it in [the `README`][readme] and in `presubmit.yml` files from the `.bcr` folder
  - [x] add a note about this change to the [`CHANGELOG`][changelog]
- [x] List user facing changes in the [`CHANGELOG`][changelog] by summarising all significant
      pull requests since the last release. Specifically:
  - Add a "Highlights" section for major improvements/changes.
  - Create "Added", "Removed", "Changed" and "Fixed" sections, as necessary.
  - If relevant, add links to the corresponding PRs to the entries.
  - Look through:
    * [merged PRs](https://github.com/tweag/rules_nixpkgs/pulls?q=is:pr+base:master+merged:>2023-10-18+-author:app/github-actions+-author:app/dependabot) or
    * `git log master ^v0.x.x --oneline --merges --grep='pull request' --grep='#' | grep -v 'tweag/dependabot/github_actions/' | grep -v 'tweag/update_flake_lock_action'`
- [x] Bump version numbers in `MODULE.bazel` and the registry, rename files
    ```
        grep '0.10.0 -r --exclude-dir=.git lists occurences of 0.10.0 in files
        find -path ./.git -prune -o -name '*0.10.0*' -print lists occurences of 0.10.0 in names
    ```
- [ ] Push the `release-<major>.<minor>.<patch>` branch and open a PR,
      go through review and merge upon success.
- [ ] Trigger the `Prepare Release` workflow
  - either via the Github UI **or**
  - run `gh workflow run -f version=<major>.<minor>.<patch> 'Prepare Release'` using the Github CLI
- [ ] Go to the [releases], open the draft release which was created to inspect it
  - Do the code snippets look valid?
  - Is there a release artifact attached to it?
  - If you're happy, publish the release... :rocket:
- [ ] Announce the new version on Twitter by asking someone with access.

[changelog]: /CHANGELOG.md
[readme]: /README.md
[releases]: https://github.com/tweag/rules_nixpkgs/releases

avdv:
> - [ ] Bump version numbers in MODULE.bazel and the registry, rename files 

[...]

PR #755, opened 2026-09-02
Release 0.14.0

vreuter:
See changelog for release notes.


.bcr/core/metadata.template.json
```diff
@@ -6,11 +6,6 @@
       "github": "benradf",
       "name": "Benjamin Radford"
     },
-    {
-      "email": "andreas.herrmann@tweag.io",
-      "github": "aherrmann",
-      "name": "Andreas Herrmann"
-    },
```
avdv, line 13: Actually, no-one mentioned as a maintainer here is still available. I'm going to add myself.

.bcr/core/presubmit.yml
```diff
@@ -3,7 +3,7 @@ bcr_test_module:
   matrix:
     platform: ["ubuntu2204"]
     bazel:
-      - "6.x"
+      - "7.x"
```
avdv, line 7: It would be nice if we could keep 6.x in the BCR to be able for users to smoothly transition to a later version. If CI in the BCR fails for 6.x and cannot be fixed easily we can still drop it later.

core/MODULE.bazel
```diff
@@ -1,6 +1,6 @@
 module(
     name = "rules_nixpkgs_core",
-    version = "0.13.0",
+    version = "0.14.0",
```
avdv: can we remove the version from the modules as suggested?  https://github.com/tweag/rules_nixpkgs/issues/749#issuecomment-5452853452

this would make releasing new versions easier

avdv (changes requested):
Thank you, just a few remarks.

core/MODULE.bazel
vreuter: Sorry I'd missed your follow-up comment to #749 . Yes, this is done here: 6c17450990968a44d938f7167329c6294c6ecdfb

.bcr/core/presubmit.yml
vreuter: OK, this is updated here: 6fea60f494c37ed1d2de84545431c913f1c40fbf

.bcr/core/metadata.template.json
vreuter: OK, I leave this one to you :)

avdv (approved):
Thank you!

PR #636, opened 2025-01-04
Update rules_nodejs version

dimatosaurus:

PR #756, opened 2026-09-02
Update rules_nodejs and resolve bug in nodejs toolchain

tshaynik:
Update rules_nodejs version to 6.7.5

This supersedes #636 and targets the newest rules_nodejs version. Adding the missing npm filegroup and toolchain attribute also resolves a bug.


tshaynik:
Important to note that #677 also accomplished some of #636, but was just missing the filegroup and the attribute.

SamirTalwar (approved):

PR #761, opened 2026-09-03
fix(util): If stat fails, assume the file is not executable.

SamirTalwar:
On failure, `stdout` is typically empty, and `int(stdout, 8)` will raise an exception, causing unclear output in the Bazel build.

This typically happens if e.g. the file does not exist. If we return `False`, Bazel correctly detects the file is not present and reports a useful error message.


avdv (approved):
Thanks!

PR #762, opened 2026-09-07
chore: switch to the publish-to-BCR workflow

vreuter:
This is needed because the publish-to-BCR app is being discontinued. See this [issue comment](https://github.com/tweag/rules_nixpkgs/issues/749#issuecomment-5567780968) by @avdv

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,32 @@
+name: Publish to Bazel Central Registry (BCR)
+on:
+  workflow_call:
+    inputs:
+      tag_name:
+        required: true
+        type: string
+    secrets:
+      publish_token:
+        required: true
+
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'Version to release (e.g. 0.11.0)'
+        required: true
+        type: string
+
+jobs:
+  publish:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.5.0
+    with:
+      tag_name: ${{ inputs.tag_name || github.ref_name }}
+      registry_fork: tweag/bazel-central-registry
+      attest: false
+      draft: false
+    permissions:
+      attestations: write
+      contents: write
+      id-token: write
+    secrets:
+      publish_token: ${{ secrets.publish_token || secrets.BCR_PUBLISH_TOKEN }}
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
     inputs:
       tag_name:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -49,3 +49,16 @@ jobs:
             --title v${{ inputs.version }} \
             v${{ inputs.version }} \
             rules_nixpkgs-${{ inputs.version }}.tar.gz
+  publish:
+    name: Publish to BCR
+    if: startsWith(github.ref, 'refs/tags/v')
```
avdv: I think this condition is unnecessary / not applicable. This workflow was not triggered by a tag and we already checked the `version` input in the release job.
```suggestion
```

.github/workflows/release.yaml
```diff
@@ -49,3 +49,16 @@ jobs:
...
+    needs: release
+    uses: ./.github/workflows/publish.yaml
+    with:
+      tag_name: ${{ inputs.version }}
```
avdv: This needs the `v` prefix.
```suggestion
      tag_name: v${{ inputs.version }}
```

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,32 @@
...
+        required: true
+        type: string
+    secrets:
+      publish_token:
+        required: true
+
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'Version to release (e.g. 0.11.0)'
```
avdv: This input should actually be a tag name.

```suggestion
        description: 'Name of the tag to release (e.g. v0.11.0)'
```

Alternatively, we could accept a `version` input here (more in line with the release workflow) and later use `${{ inputs.tag_name || (inputs.version && format('v{0}', inputs.version)) }}` but that seems overly complicated.

avdv (changes requested):
Thank you, just some minor remarks.

.github/workflows/publish.yaml
vreuter: I went with your suggestion here: https://github.com/tweag/rules_nixpkgs/pull/762/changes/f59d87286a577172febcd5823eff3b82a77582c6

.github/workflows/release.yaml
vreuter: I went with your suggestion here: https://github.com/tweag/rules_nixpkgs/pull/762/changes/f59d87286a577172febcd5823eff3b82a77582c6

.github/workflows/release.yaml
vreuter: I went with your suggestion here: https://github.com/tweag/rules_nixpkgs/pull/762/changes/f59d87286a577172febcd5823eff3b82a77582c6

avdv (approved):
Thank you!

